PR #1, opened 2015-11-03
Implemented the FreeTextResponse XBlock

azimpradhan:
Many instructors want a problem type that allows for free text
responses. Previously, they were made to cut and paste javascript
into the Blank Common Problem and provide a regular expression that
checks a student's response for the correct answer. The FreeText
XBlock allows for instructors to create free text response questions
in their course. The following features are supported by the XBlock:
-Assigning grades to students
-Customizing the prompt for the student response
-Enforcing a minimum or maximum word count
-Setting a list of key phrases that earn half-credit
-Setting a list of key phrases that earn full-credit
-Customize whether to display the correctness indicator to students

A suite of tests have been written for this XBlock and can be run
as follows:
python ./manage.py lms test --verbosity=1
freetextresponse/tests.py:FreetextResponseXblockTests --traceback
--settings=test


azimpradhan:
@dcadams, @caesar2164, please review :)

freetextresponse/private/edit.js
```diff
@@ -1,18 +1,19 @@
...
-    'use strict';
-
+function FreetextresponseEdit(runtime, element) {
     var $ = window.$;
     var $element = $(element);
     var buttonSave = $element.find('.save-button');
     var buttonCancel = $element.find('.cancel-button');
     var url = runtime.handlerUrl(element, 'studio_view_save');
 
+
```
caesar2164: maybe remove?

freetextresponse/private/edit.js
```diff
@@ -1,18 +1,19 @@
...
     buttonCancel.on('click', function () {
         runtime.notify('cancel', {});
+        console.log('cancel button was pressed!');
```
caesar2164: remove test logs before merging

freetextresponse/private/edit.js
```diff
@@ -1,18 +1,19 @@
...
         return false;
     });
 
     buttonSave.on('click', function () {
+        console.log('save button was pressed!');
```
caesar2164: remove test logs before merging

freetextresponse/private/view.html
```diff
@@ -0,0 +1,21 @@
+<div class="freetextresponse_block">
+    <h2 class="problem-header">{self.display_name}</h2>
+    <div class="problem-progress">({problem_progress})</div>
+    <h3>{self.prompt}</h3>
+    <div class="{indicator_class}">
```
caesar2164: add class like "correctness-indicator"

freetextresponse/private/view.less
```diff
@@ -0,0 +1,77 @@
...
+    height: 20px;
+    background: url("images/incorrect-icon.png") center center no-repeat;
+  }
+
+  div p.status {
+    margin: 8px 0 0 10px;
+    text-indent: 100%;
+    white-space: nowrap;
+    overflow: hidden;
+  }
```
caesar2164, line 27: with added class from view.html lines 4-34 can be rewritten as:

```
.correctness-indicator {
  .status{
       margin-left: 10px;
  }
  &.unanswered .status {
    ...
  }
  &.correct .status {
    ...
  }
  &.incorrect .status {
    ...
  }
}
```

freetextresponse/private/view.less
```diff
@@ -0,0 +1,77 @@
...
+
+  .incorrect span.status {
+    display: inline-block;
+    width: 20px;
+    height: 20px;
+    background: url("images/incorrect-icon.png") center center no-repeat;
+  }
+
+  div p.status {
```
caesar2164: remove this CSS block.

freetextresponse/private/view.less
```diff
@@ -0,0 +1,77 @@
...
+    margin-top: 8px;
+    margin-left: 10px;
+    color: #666;
+    font-style: italic;
+    -webkit-font-smoothing: antialiased;
+  }
+
+  .action .nodisplay{
+    display: none;
+  }
```
caesar2164: refactor 36-65 as:

```
.action {
  .check {
  }
  .used_attempts_feedback {
  }
  .nodisplay{
  }
}
.problem-progress {
}
```

freetextresponse/public/edit.js
```diff
@@ -1,18 +1,19 @@
...
-    'use strict';
-
+function FreetextresponseEdit(runtime, element) {
     var $ = window.$;
     var $element = $(element);
     var buttonSave = $element.find('.save-button');
     var buttonCancel = $element.find('.cancel-button');
     var url = runtime.handlerUrl(element, 'studio_view_save');
 
+
```
caesar2164: remove

freetextresponse/public/edit.js
```diff
@@ -1,18 +1,19 @@
...
     buttonCancel.on('click', function () {
         runtime.notify('cancel', {});
+        console.log('cancel button was pressed!');
```
caesar2164: remove

freetextresponse/public/edit.js
```diff
@@ -1,18 +1,19 @@
...
         return false;
     });
 
     buttonSave.on('click', function () {
+        console.log('save button was pressed!');
```
caesar2164: remove

freetextresponse/public/edit.html
```diff
@@ -0,0 +1 @@
...ng-label" for="freetextresponse_fullcredit_keyphrases">Full-Credit Keyphrases</label><input class="input setting-input" id="freetextresponse_fullcredit_keyphrases" value="{self.fullcredit_keyphrases}" tabindex="1"></div><span class="tip setting-help">This is a comma separated list of keyphrases, that if present in the student's response, will assign the student full credit.</span></li><li class="field comp-setting-entry is-set"><div class="wrapper-comp-setting"><label class="label setting-label" for="freetextresponse_halfcredit_keyphrases">Half-Credit Keyphrases</label><input class="input setting-input" id="freetextresponse_halfcredit_keyphrases" value="{self.halfcredit_keyphrases}" tabindex="1"></div><span class="tip setting-help">This is a comma separated list of keyphrases, that if present in the student's response, will assign the student half credit, assuming the response does not already contain a fullcredit keyphrase.</span></li></ul><div class="xblock-actions"><ul><li class="action-item"><a href="#" tabindex="2" class="button action-primary save-button">Save</a></li><li class="action-item"><a href="" tabindex="3" class="button cancel-button">Cancel</a></li></ul></div></div>
```
caesar2164: add newline at end of file

caesar2164: nevermind, this is autogen from private/edit.html

freetextresponse/public/view.js
```diff
@@ -0,0 +1,47 @@
...
+    var buttonSubmit = $element.find('.check.Submit');
+    var textareaParent = $element.find('.student_answer').parent();
+    var usedAttemptsFeedback = $element.find('.action .used_attempts_feedback');
+    var problemProgress = $element.find('.problem-progress');
+    var submitParent = $element.find('.Submit').parent();
+    var wordCountError = $element.find('.word-count-error');
+    var url = runtime.handlerUrl(element, 'submit');
+
+    buttonSubmit.on('click', function () {
+        console.log('submit button was pressed!');
```
caesar2164: remove this line

freetextresponse/tests.py
```diff
@@ -0,0 +1,512 @@
...
+        course_id = SlashSeparatedCourseKey('foo', 'bar', 'baz')
+        runtime = mock.Mock(course_id=course_id)
+        scope_ids = mock.Mock()
+
+        field_data = DictFieldData(kw)
+        xblock = Freetextresponse(runtime, field_data, scope_ids)
+        xblock.xmodule_runtime = runtime
+        return xblock
+
+    def setUp(self):
```
caesar2164: two lines before new definitions

caesar2164:
Looking good!

README.markdown
```diff
@@ -4,22 +4,22 @@ Enables instructors to create questions with free-text responses.
 ## TODO List:
```
dcadams: Is the TODO list still relevant?

dcadams: It would be helpful to include some instructions for how to set this up in Studio e.g. what do I put in the Advanced Modules list.

freetextresponse/public/edit.js
```diff
@@ -21,7 +22,15 @@ function XblockFreetextEdit(runtime, element) {
             type: 'POST',
             data: JSON.stringify({
                 // TODO: Add entries here for each field to be saved
```
dcadams: remove

freetextresponse/freetextresponse.py
```diff
@@ -0,0 +1,437 @@
...
+        ]
+    display_name = String(
+        default='FreeText XBlock',
+        scope=Scope.settings,
+        help="This is the title for this question type",
+    )
+    prompt = String(
+        default='Please enter your response within this text area',
+        scope=Scope.settings,
+        help='This is the prompt students will see when '
```
dcadams: Does it use this help or the one defined in edit.html? I suspect it is using the edit.html one but the xblock docs specify it like this however.

freetextresponse/freetextresponse.py
```diff
@@ -0,0 +1,437 @@
...
+from enum import Enum
+from xblock.core import XBlock
+from xblock.fields import Scope
+from xblock.fields import String
+from xblock.fields import Integer
+from xblock.fields import Float
+from xblock.fragment import Fragment
+
+
+class Freetextresponse(XBlock):
```
stvstnfrd: `FreeTextResponse`

freetextresponse/freetextresponse.py
```diff
@@ -0,0 +1,437 @@
...
+    #  pylint: disable=too-many-ancestors, too-many-instance-attributes
+    '''
```
stvstnfrd: Per the style guide, docstrings should always use double-quotes.

freetextresponse/tests.py
```diff
@@ -0,0 +1,512 @@
...
+
+    def test_used_attempts_feedback_normal(self):
+        # pylint: disable=invalid-name
+        '''
+        Tests that get_used_attempts_feedback returns the expected feedback
+        '''
+        self.xblock.max_attempts = 5
+        self.xblock.num_attempts = 3
+        self.assertEquals(
+            'You have used 3 of 5 submissions',
```
stvstnfrd: Should this, and other text, be i18n'ed?

freetextresponse/tests.py
```diff
@@ -0,0 +1,512 @@
...
+        Tests that determine_credit() returns full-credit when appropriate
+        '''
+        self.xblock.student_answer = 'Non-blank answer'
+        self.xblock.word_count_valid = MagicMock(return_value=True)
+        self.xblock.fullcredit_keyphrases = 'Non-blank, answer'
+        self.xblock.halfcredit_keyphrases = 'Something else'
+        self.assertEquals(Credit.full, self.xblock.determine_credit())
+
+    def test_used_attempts_feedback_blank(self):
+        # pylint: disable=invalid-name
```
stvstnfrd: Here and throughout, which names are invalid?

freetextresponse/freetextresponse.py
stvstnfrd: Related: We should look into the work they've done in the xblock-utils library [1], namely creating a default studio view [2].

My work (here: the default studio view) predates the creation of `xblock-utils`. Despite having not explored it thoroughly, it does appear to be a superior implementation.

[edit]...

Upon quick inspection, it _does_ appear to handle help text automagically [3].

Given all of that, we should look to adopt their approach and use their library.

[1] https://github.com/edx/xblock-utils
[2] https://github.com/edx/xblock-utils/blob/08e5a5a9fc8ab46b627435427fd7e04c20809009/xblockutils/studio_editable.py#L69
[3] https://github.com/edx/xblock-utils/blob/08e5a5a9fc8ab46b627435427fd7e04c20809009/xblockutils/studio_editable.py#L117

README.markdown
stvstnfrd: Yeah, the checklist is just meant to be a "mini guide" during development; we want to update it prior to merge.

freetextresponse/__init__.py
```diff
@@ -0,0 +1,6 @@
+'''
+This is an XBlock that accepts a free-text response from students.
+Instructors can specify a list of phrases, of which one must be
+present in order for the student to receive credit.
+'''
+from .freetextresponse import Freetextresponse
```
stvstnfrd: I wonder if we need this or if it's mostly cargo cult at this point..?

Correspondingly, we may need to update the `entry_points` in `setup.py` (if we remove this line here:

``` python
entry_points={
    'xblock.v1': [
        'freetextresponse = freetextresponse.freetextresponse:Freetextresponse',
    ],
},
```

Mind giving it a try?

freetextresponse/freetextresponse.py
```diff
@@ -0,0 +1,437 @@
...
+from xblock.core import XBlock
+from xblock.fields import Scope
+from xblock.fields import String
+from xblock.fields import Integer
+from xblock.fields import Float
+from xblock.fragment import Fragment
+
+
+class Freetextresponse(XBlock):
+    #  pylint: disable=too-many-ancestors, too-many-instance-attributes
```
stvstnfrd, line 24: Where does the `too-many-ancestors` error come from, the classes from which `XBlock` is derived?

freetextresponse/freetextresponse.py
```diff
@@ -0,0 +1,437 @@
...
+        return [
+            ('FreeText XBlock',
+             '''<sequence_demo>
+                    <freetextresponse />
+                    <freetextresponse name="My First XBlock" />
+                </sequence_demo>
+             '''),
+        ]
+    display_name = String(
+        default='FreeText XBlock',
```
stvstnfrd: i18n for all user-facing strings?

freetextresponse/freetextresponse.py
```diff
@@ -0,0 +1,437 @@
...
+        scope=Scope.user_state,
+        help="This is the score of the user",
+    )
+    max_attempts = Integer(
+        default=0,
+        scope=Scope.settings,
+        help='This is the maximum number of times a student '
+             'is allowed to attempt the problem',
+    )
+    display_correctness = String(
```
stvstnfrd: Shouldn't this be a `bool`?

freetextresponse/freetextresponse.py
```diff
@@ -0,0 +1,437 @@
...
+             'icon should be displayed after a student enters '
+             'their response',
+    )
+    min_word_count = Integer(
+        default=0,
+        scope=Scope.settings,
+        help='This is the minimum number of words required for this question',
+    )
+    max_word_count = Integer(
+        default=10000,
```
stvstnfrd: Would `None` be more apt?

freetextresponse/freetextresponse.py
```diff
@@ -0,0 +1,437 @@
...
+            paths_js=[
+                'edit.js.min.js',
+            ],
+            fragment_js='FreetextresponseEdit',
+        )
+        return fragment
+
+    @XBlock.json_handler
+    def studio_view_save(self, data, suffix=''):
+        #  pylint: disable=unused-argument
```
stvstnfrd: Elsewhere, you add this pragma on the same line as the `def`; which approach should we use consistently?

Gruntfile.js
```diff
@@ -3,7 +3,7 @@ module.exports = function (grunt) {
 
     var jshintrc = '.jshintrc';
     var gruntFile = 'Gruntfile.js';
-    var directoryPackage = './xblockfreetext';
+    var directoryPackage = './freetextresponse';
```
stvstnfrd, line 6: I much prefer your naming convention here.

freetextresponse/freetextresponse.py
```diff
@@ -0,0 +1,437 @@
...
+    @staticmethod
+    def workbench_scenarios():
+        '''
+        Gather scenarios to be displayed in the workbench
+        '''
+        return [
+            ('FreeText XBlock',
+             '''<sequence_demo>
+                    <freetextresponse />
+                    <freetextresponse name="My First XBlock" />
```
stvstnfrd: Can we add more illustrative example instances here?

stvstnfrd: It would be great to include a spectrum of configurations here that demonstrate at least one example of each kind of "check".

freetextresponse/freetextresponse.py
stvstnfrd: Just a thought, is there maybe a better default?

At the least, I think we should change "XBlock"; instructors don't care if it's an XBlock/XModule/CAPA-problem/LTI-component/etc. and students will have no idea. Maybe something like "Free-text Response" is suitably more general?

Ideally, instructors would give everything an apt display name, but realistically many won't. This way, our default case is less cryptic to non-developers.

freetextresponse/freetextresponse.py
```diff
@@ -0,0 +1,437 @@
...
+                </sequence_demo>
+             '''),
+        ]
+    display_name = String(
+        default='FreeText XBlock',
+        scope=Scope.settings,
+        help="This is the title for this question type",
```
stvstnfrd: Here and throughout, single quotes for non-interpolated strings.

freetextresponse/freetextresponse.py
```diff
@@ -0,0 +1,437 @@
...
+        default=0,
+        scope=Scope.settings,
+        help='This is the minimum number of words required for this question',
+    )
+    max_word_count = Integer(
+        default=10000,
+        scope=Scope.settings,
+        help='This is the maximum number of words allowed for this question',
+    )
+    fullcredit_keyphrases = String(
```
stvstnfrd: Should this be a `List`?

freetextresponse/freetextresponse.py
```diff
@@ -0,0 +1,437 @@
...
+        default='Full-Credit Phrase 1, Full-Credit Phrase 2, '
+                'Full-Credit Phrase 3',
+        scope=Scope.settings,
+        help="This is a comma-separated list of words or phrases, any of "
+             "which must be present in order for the student's answer "
+             "to receive full credit",
+    )
+    halfcredit_keyphrases = String(
```
stvstnfrd: By storing these as a list, we don't tie field values on the back-end to an implementation detail on the front-end (stored as comma-separated values).

freetextresponse/freetextresponse.py
```diff
@@ -0,0 +1,437 @@
...
+        default='Half-Credit Phrase 1, Half-Credit Phrase 2, '
+                'Half-Credit Phrase 3',
+        scope=Scope.settings,
+        help="This is a comma-separated list of words or phrases, any of "
+             "which must be present in order for the student's answer "
+             "to receive half credit",
+    )
+    student_answer = String(
+        default='',
```
stvstnfrd: What if the default was `None`?

The distinction being that `''` (the empty string) indicates a response with no content, while `None` (the empty set) indicates no response at all.

freetextresponse/freetextresponse.py
```diff
@@ -0,0 +1,437 @@
...
+        scope=Scope.user_state,
+        help="This is the body of the user's free-text answer",
+    )
+    num_attempts = Integer(
```
stvstnfrd: Perhaps `count_attempts` is more descriptive? It indicates not just that this field is a number of something, but that it is a tally of occurrences of some event.

freetextresponse/freetextresponse.py
```diff
@@ -0,0 +1,437 @@
...
+        return fragment
+
+    @XBlock.json_handler
+    def studio_view_save(self, data, suffix=''):
+        #  pylint: disable=unused-argument
+        '''
+        Save XBlock fields
+        Returns: the new field values
+        '''
+        return self.save_studio_data(data)
```
stvstnfrd: Why the wrapper method if there's no additional logic?

Any reason not to just move the body of `save_studio_data` directly into `studio_view_save`?

freetextresponse/freetextresponse.py
```diff
@@ -0,0 +1,437 @@
...
+
+    def save_studio_data(self, data):
+        '''
+        Helper method to save the parameters set by the instructor
+        for the given instance of the FreeTextResponse XBlock
+        '''
+        self.display_name = data['display_name']
+        self.prompt = data['prompt']
+        if data['weight'] and int(float(data['weight'])) >= 0:
```
stvstnfrd: What happens if I try to save a value of `"fifty"` in the `weight` field?

stvstnfrd: Would this break if the AJAX call didn't supply the `weight` parameter?

freetextresponse/freetextresponse.py
```diff
@@ -0,0 +1,437 @@
...
+            self.weight = int(float(data['weight']))
```
stvstnfrd: I hate to see this (double) cast performed twice in a row.

What about:

``` python
if 'weight' in data:
    try:
        weight = int(float(data['weight']))
    except ValueError:
        pass
    else:
        self.weight = weight
```

stvstnfrd: Also, why the double cast?

freetextresponse/freetextresponse.py
```diff
@@ -0,0 +1,437 @@
...
+        if data['max_attempts'] and int(float(data['max_attempts'])) >= 0:
```
stvstnfrd: See comments just above ^

freetextresponse/freetextresponse.py
```diff
@@ -0,0 +1,437 @@
...
+            self.max_attempts = int(float(data['max_attempts']))
+        self.display_correctness = data['display_correctness']
```
stvstnfrd: Should this be

``` python
if 'display_correctness' in data:
    self.display_correctness = data['display_correctness']
```

That would allow a front-end to omit the parameter if it were unchanged.

freetextresponse/freetextresponse.py
```diff
@@ -0,0 +1,437 @@
...
+        if data['min_word_count'] and data['max_word_count']:
+            if (
```
stvstnfrd: See note above about exception handling these casts.

Can we memoize these?

``` python
if 'min_word_count' in data and 'max_word_count' in data:
    min_word_count = int(float(data['min_word_count']))
    max_word_count = int(float(data['max_word_count']))
    if min_word_count >=0 and min_word_count <= max_word_count:
        # ...
```

freetextresponse/freetextresponse.py
```diff
@@ -0,0 +1,437 @@
...
+                self.min_word_count = int(float(data['min_word_count']))
+            if (
+                    int(float(data['max_word_count'])) >= 0 and
+                    (
+                        int(float(data['max_word_count'])) >=
+                        int(float(data['min_word_count']))
+                    )
+            ):
+                self.max_word_count = int(float(data['max_word_count']))
+        self.fullcredit_keyphrases = data['fullcredit_keyphrases']
```
stvstnfrd: See note above about storing this as a list. We could clean it up with something like:

``` python
if 'full_credit_key_phrases' in data:
    self.full_credit_key_phrases = [
        phrase.strip()
        for phrase in data['full_credit_key_phrases'].split(',')
        if phrase.strip()
    ]
```

freetextresponse/freetextresponse.py
```diff
@@ -0,0 +1,437 @@
...
+        self.halfcredit_keyphrases = data['halfcredit_keyphrases']
```
stvstnfrd: Whatever we decide with `full_credit_key_phrases`, we should do similarly here.

freetextresponse/freetextresponse.py
```diff
@@ -0,0 +1,437 @@
...
+        return ''
+
+    def get_word_count_message(self):
+        '''
+        Returns the word count message based on the student's answer
+        '''
+        if self.num_attempts < 1 or self.word_count_valid():
+            return ''
+        else:
+            return "Invalid Word Count. Your reponse must be " \
```
stvstnfrd: One reason I prefer a single return statement is that when debugging, you can set a breakpoint on the one `return` statement and run straight there. Otherwise, you have to step a couple of times. Jumping right to a single return can make it immediately clear what's going on (here: if the result is non-empty, we know the `if` check was `False`. With multiple `return` statements like this, you have to take at least one `step` through on your debugger after hitting that first break point. :\

stvstnfrd: Also, we should generally avoid making function calls from within `return` statements (use `return "value"` instead of `return "value".do_stuff()`. It's a pain when stepping through on your debugger, being returned immediately without ever being stored in a variable.

freetextresponse/freetextresponse.py
```diff
@@ -0,0 +1,437 @@
...
+        for url in urls_js:
+            fragment.add_javascript_url(url)
+        for path in paths_js:
+            url = self.get_resource_url(path)
+            fragment.add_javascript_url(url)
+        if fragment_js:
+            fragment.initialize_js(fragment_js)
+        return fragment
+
+    def get_indicator_visiblity_class(self):
```
stvstnfrd: Are these methods intended to be private/local? If so, can we indicate so by prefixing the names with an underscore? `def _get_indicator_visibility_class(self):`

freetextresponse/freetextresponse.py
```diff
@@ -0,0 +1,437 @@
...
+                .format(
+                    min=self.min_word_count,
+                    max=self.max_word_count,
+                )
+
+    def get_indicator_class(self):
+        '''
+        Returns the class of the correctness indicator element
+        '''
+        if self.student_answer == '':
```
stvstnfrd: If we default `student_answer` to `None`, per above, this can be simplified like:

``` python
if self.student_answer is None:
    return 'unanswered'
elif not self.student_answer:
    return 'incorrect'
elif  ( # ...
```

freetextresponse/freetextresponse.py
```diff
@@ -0,0 +1,437 @@
...
+    max_word_count = Integer(
+        default=10000,
+        scope=Scope.settings,
+        help='This is the maximum number of words allowed for this question',
+    )
+    fullcredit_keyphrases = String(
+        default='Full-Credit Phrase 1, Full-Credit Phrase 2, '
+                'Full-Credit Phrase 3',
+        scope=Scope.settings,
+        help="This is a comma-separated list of words or phrases, any of "
```
stvstnfrd: "one of which" ?

freetextresponse/freetextresponse.py
```diff
@@ -0,0 +1,437 @@
...
+        )
+
+    def phrase_present_in_answer(self, phrase_list):
+        '''
+        Helper method that returns a boolean value indicating
+        whether at least one member of the provided phrase list
+        is present in the student's answer
+        '''
+        if re.search(
+                '|'.join(
```
stvstnfrd: Dynamically generated regexs; eek!

In this context, I'm not sure regexs are the best tool for the task.
They're not super efficient and this could be a pain to debug;
what happens if my phrase contains a pipe `|` or some other special chars?

A more pythonic approach would be something like

``` python
def is_at_least_one_phrase_present(phrases, answer):
    answer = answer.lower()
    matches = [
        phrase.lower() in answer
        for phrase in phrases
    ]
    return any(matches)
```

or 

``` python
if any([phrase.lower() in answer.lower() for phrase in phrases]):
    # ...
```

... if you're feeling clever [1]

;)

[1] Clever code is a pain to maintain; don't be so clever ;)

freetextresponse/freetextresponse.py
```diff
@@ -0,0 +1,437 @@
...
+        return False
+
+    def get_problem_progress(self):
+        '''
+        Returns a statement of progress for the XBlock, which depends
+        on the user's current score
+        '''
+        if self.score == 0.0:
+            return (
+                str(self.weight)
```
stvstnfrd: `unicode`?

freetextresponse/freetextresponse.py
```diff
@@ -0,0 +1,437 @@
...
+                + ' point'
+                + ('s' if (self.weight > 1) else '')
```
stvstnfrd: Single line `if` statements are bad [1]; single line ternary statements are worse :)

[1] Where "bad" means "not immediately obvious" and "requires additional mental parsing than a multiline if/else statement."

freetextresponse/freetextresponse.py
```diff
@@ -0,0 +1,437 @@
...
+                + ' possible'
+            )
+        else:
+            return (
+                str(self.score).replace('.0', '')
```
stvstnfrd: This string formatting is wonky; let's just use `.format()`.

freetextresponse/freetextresponse.py
```diff
@@ -0,0 +1,437 @@
...
+        '''
+        credit = self.determine_credit()
+        if credit == Credit.full:
+            self.score = self.weight
+        elif credit == Credit.half:
+            self.score = float(self.weight)/2
+        else:
+            self.score = 0.0
+        self.runtime.publish(
+            self, "grade",
```
stvstnfrd: Drop `'grade'` down to a new line.

freetextresponse/freetextresponse.py
```diff
@@ -0,0 +1,437 @@
...
+    def compute_score(self):
+        '''
+        Computes and publishes the user's core for the XBlock
+        based on their answer
+        '''
+        credit = self.determine_credit()
+        if credit == Credit.full:
+            self.score = self.weight
+        elif credit == Credit.half:
+            self.score = float(self.weight)/2
```
stvstnfrd, line 372: Could we simply ensure the cast with `self.score = self.weight / 2.0`?

freetextresponse/freetextresponse.py
```diff
@@ -0,0 +1,437 @@
...
+        )
+
+    def determine_credit(self):
+        '''
+        Helper Method that determines the level of credit that
+        the user should earn based on their answer
+        '''
+        if self.student_answer == '' or not self.word_count_valid():
+            return Credit.zero
+        if self.phrase_present_in_answer(self.fullcredit_keyphrases):
```
stvstnfrd: This is oddly inconsistent, switching from `if`->`return` -> `if`->`return` to `if`->`return`->`elif`->...

At the least, we should chose one approach and use it consistently.

I prefer independent blocks like:

``` python
if True:
    return True
if False:
    return False
return None
```

This allows you to remove individual blocks without having to rewrite logic (like when you delete the first `if` check in a multi-clause `if`-`elif` statement.

freetextresponse/freetextresponse.py
```diff
@@ -0,0 +1,437 @@
...
+
+    def get_used_attempts_feedback(self):
+        '''
+        Returns the text with feedback to the user about the number of attempts
+        they have used if applicable
+        '''
+        if self.max_attempts == 0:
+            return ''
+        return (
+            'You have used '
```
stvstnfrd: Not only should this be `i18n`ed, it should use `ngettext` to handle pluralization.

freetextresponse/private/edit.html
```diff
@@ -0,0 +1,174 @@
+<div class="wrapper-comp-settings is-active editor-with-buttons xblockfreetext_edit" id="settings-tab">
+  <ul class="list-input settings-list">
+    <!-- TODO: Add an entry here for each field -->
```
stvstnfrd: Remove note when TODONE ;)

stvstnfrd: Actually, see note above about using `xblock-utils` instead here.. We can discuss further.

freetextresponse/private/edit.js
```diff
@@ -1,18 +1,19 @@
-function XblockFreetextEdit(runtime, element) {
-    'use strict';
-
+function FreetextresponseEdit(runtime, element) {
```
stvstnfrd: `FreeTextResponseEdit`?

freetextresponse/private/edit.less
```diff
@@ -1,3 +1,3 @@
-.xblockfreetext_edit {
+.freetextresponse_edit {
```
stvstnfrd: If we aren't using this file, we can remove it here, as well as any references elsewhere.

freetextresponse/private/view.js
```diff
@@ -0,0 +1,47 @@
+function FreetextresponseView(runtime, element) {
```
stvstnfrd: `FreeTextResponseView`?

freetextresponse/private/view.js
```diff
@@ -0,0 +1,47 @@
...
+    var buttonSubmit = $element.find('.check.Submit');
+    var textareaParent = $element.find('.student_answer').parent();
+    var usedAttemptsFeedback = $element.find('.action .used_attempts_feedback');
+    var problemProgress = $element.find('.problem-progress');
+    var submitParent = $element.find('.Submit').parent();
+    var wordCountError = $element.find('.word-count-error');
+    var url = runtime.handlerUrl(element, 'submit');
+
+    buttonSubmit.on('click', function () {
+        console.log('submit button was pressed!');
```
stvstnfrd: Drop debug statements.

freetextresponse/private/view.js
```diff
@@ -0,0 +1,47 @@
...
+        buttonSubmit.text('Checking...');
+        runtime.notify('submit', {
+            message: 'Submitting...',
+            state: 'start',
```
stvstnfrd: Just to point it out... since this JS gets compiled, we can include trailing commas in Javascript (still not in JSON, though).

freetextresponse/private/view.less
```diff
@@ -0,0 +1,77 @@
+.freetextresponse_block {
+    // TODO: Put your styles here
```
stvstnfrd: Remove when TODONE ;)

freetextresponse/private/view.less
```diff
@@ -0,0 +1,77 @@
+.freetextresponse_block {
```
stvstnfrd: I also think the `_block` suffix is cargo cult and could probably be dropped.

Thoughts?

freetextresponse/private/view.less
```diff
@@ -0,0 +1,77 @@
...
+    font-size: 1em;
+  }
+
+  .action .used_attempts_feedback {
+    display: inline-block;
+    margin-top: 8px;
+    margin-left: 10px;
+    color: #666;
+    font-style: italic;
+    -webkit-font-smoothing: antialiased;
```
stvstnfrd: What about non-webkit browsers?

freetextresponse/private/view.less
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  .action .nodisplay{
+    display: none;
+  }
+
+  .hidden{
```
stvstnfrd, line 61: Add a space before the `{`.

freetextresponse/private/view.less
```diff
@@ -0,0 +1,77 @@
...
+    color: #666;
+    font-weight: 100;
+    font-size: 1em;
+  }
+
+  .action .used_attempts_feedback {
+    display: inline-block;
+    margin-top: 8px;
+    margin-left: 10px;
+    color: #666;
```
stvstnfrd: We use this color more than once; should it be a variable?

freetextresponse/tests.py
```diff
@@ -0,0 +1,512 @@
...
+import mock
+from mock import MagicMock
+from django.test.client import Client
+from opaque_keys.edx.locations import SlashSeparatedCourseKey
+from xblock.field_data import DictFieldData
+from .freetextresponse import Freetextresponse
+from .freetextresponse import Credit
+
+
+class FreetextResponseXblockTests(unittest.TestCase):
```
stvstnfrd: `class FreeTextResponseXblockTestCase(unittest.TestCase):`

freetextresponse/tests.py
```diff
@@ -0,0 +1,512 @@
...
+    '''
+    @classmethod
+    def make_an_xblock(cls, **kw):
+        '''
+        Helper method that creates a FreeTextResponse XBlock
+        '''
+        course_id = SlashSeparatedCourseKey('foo', 'bar', 'baz')
+        runtime = mock.Mock(course_id=course_id)
+        scope_ids = mock.Mock()
+
```
stvstnfrd: Nit: remove newline.

freetextresponse/__init__.py
azimpradhan: This line seems to be necessary. Tried what you suggested, but it didn't work

freetextresponse/freetextresponse.py
azimpradhan, line 24: yes

freetextresponse/freetextresponse.py
azimpradhan: ideally yes, but because our xblock html files can only render strings from python, display_correctness is left as a string

freetextresponse/freetextresponse.py
azimpradhan: While none, would allow users to enter a response that is infinitely long, i thought putting a reasonable default value would help prevent malicious users from spamming our storage

freetextresponse/freetextresponse.py
azimpradhan: I would prefer to keep the default as `''` because helper methods depend on student_answer not being null, and it would be cumbersome to add a special case for this.

freetextresponse/freetextresponse.py
azimpradhan: the reason this is on a new line is because pep8 complained about the line being too long

freetextresponse/freetextresponse.py
azimpradhan: yes. This has been separated into a wrapper for testing purposes. The unit tests fail to execute if there is a decorator Xblock.json_handler decorator on the method

freetextresponse/freetextresponse.py
azimpradhan: You cannot enter the value of "fifty" input type is set to 'number' in the html. if the weight parameter is not supplied it is not updated

freetextresponse/freetextresponse.py
azimpradhan, line 372: Effectively, it is doing the same thing. is it not?

freetextresponse/tests.py
azimpradhan: ones that pylint thinks are too long

freetextresponse/freetextresponse.py
azimpradhan: The reason for this is because the html input of type 'number' allows the user to enter decimal points. However it is xblock convention to store the weight as a whole number.

freetextresponse/__init__.py
stvstnfrd: +1

freetextresponse/freetextresponse.py
stvstnfrd, line 24: +1

freetextresponse/freetextresponse.py
stvstnfrd: Reasonable.

freetextresponse/freetextresponse.py
azimpradhan: +1

freetextresponse/freetextresponse.py
azimpradhan: As we discussed, and you will try different scenarios on your workbench, since my workbench doesn't display the xblock

freetextresponse/freetextresponse.py
azimpradhan: +1

freetextresponse/freetextresponse.py
azimpradhan: +1

azimpradhan:
@caesar2164 @dcadams @stvstnfrd please verify that I have addressed your feedback

freetextresponse/private/view.html
```diff
@@ -0,0 +1,23 @@
+<div class="freetextresponse">
+    <h2 class="problem-header">{self.display_name}</h2>
+    <div class="problem-progress">({problem_progress})</div>
+    <h3>{self.prompt}</h3>
+    <span class="user_input">
+        <div class="{indicator_class}">
```
caesar2164: make this: `<div class="user-input {indicator_class}">`

freetextresponse/private/view.html
```diff
@@ -0,0 +1,23 @@
+<div class="freetextresponse">
+    <h2 class="problem-header">{self.display_name}</h2>
+    <div class="problem-progress">({problem_progress})</div>
+    <h3>{self.prompt}</h3>
+    <span class="user_input">
```
caesar2164: remove this wrapping span

freetextresponse/freetextresponse.py
stvstnfrd: Ugh, that's a shame. I'm fine leaving this as-is.

freetextresponse/private/edit.js
```diff
@@ -1,6 +1,4 @@
-function XblockFreetextEdit(runtime, element) {
-    'use strict';
```
stvstnfrd: Did you remove this intentionally?

stvstnfrd: Oh, I said I wouldn't comment on `studio_view` code..

README.markdown
```diff
@@ -1,33 +1,3 @@
...
-- [ ] Update package metadata
-    - [ ] `./package.json`
-        - https://www.npmjs.org/doc/files/package.json.html
-    - [ ] `./setup.py`
-        - https://docs.python.org/2/distutils/setupscript.html#additional-meta-data
-- [ ] Update `./Gruntfile.js`
-    - http://gruntjs.com/getting-started
-- [ ] Update `./README.markdown`
-- [ ] Write documentation
-- [ ] Publish on PyPi
```
stvstnfrd, line 33: I'd like to see us publishing to PyPi for all XBlocks, by default. This is pretty straight-forward; we did it for the `ImageModal` XBlock [1]. Let's sit down and discuss a plan for this.

[1] https://pypi.python.org/pypi/xblock-image-modal

caesar2164:
:+1: from me!

azimpradhan:
@dcadams, @stvstnfrd, I believe I have addressed your comments (including making use of the StudioEditableMixin). Please take a look. Thanks.

README.markdown
```diff
@@ -1,33 +1,29 @@
...
+This package provides an XBlock for use with the EdX Platform and makes
+it possible for instructors to creqte questions that expect a
```
stvstnfrd: s/creqte/create

README.markdown
```diff
@@ -1,33 +1,29 @@
...
+free-text response.
+
+Instructors define the following paramters in Studio:
+- display name
+- display correctness (True/False)
+- Full-Credit Phrases
```
stvstnfrd: Just a little more documentation here will go a long way; do you mind expanding upon what exactly "full/half-credit phrases" means?

dcadams:
lgtm, just need to squash. :+1:

README.markdown
```diff
@@ -1,33 +1,29 @@
...
+- Minimum Word Count
+- Prompt
+- Question Weight
+
+Students enter and submit their free-text responses, which instantly gets evaluated
+according to the parameters above.
+
+# Installation
+- Add the xblock to your requirements/edx/github.text file
+  e.g. -e git+https://github.com/Stanford-Online/xblock-free-text-response@cfb793db182b60281875b83b53a98640d740ebcf#egg=xblock-free-text-response
```
stvstnfrd, line 24: I don't know that we want/need to include installation instructions. While somewhat helpful, anyone stumbling across this repo should/needs to already know how XBlock installation works; it's documented elsewhere in the OpenEdX platform.

Specifically, we definitely don't want the README pinned to a specific SHA.

README.markdown
```diff
@@ -1,33 +1,29 @@
...
+
+- In Studio Settings/Advanced Settings add the xblock to the Advanced Module List.
+  e.g. "freetextresponse"
```
stvstnfrd, line 27: Though, this bit is useful, as no one should need to dig into the setup.py file just to know how to reference an installed XBlock.

freetextresponse/freetextresponse.py
```diff
@@ -0,0 +1,486 @@
+"""
+This is the core logic for the FreeText XBlock
```
stvstnfrd: Is it `FreeText` or `Free-text` or some other permutation?

While I'd lean toward `Free-text`, I don't care too strongly, though I would like to see it represented consistently throughout the code and documentation.

freetextresponse/freetextresponse.py
```diff
@@ -0,0 +1,486 @@
...
+        scope=Scope.settings,
+    )
+    min_word_count = Integer(
+        display_name=_('Minimum Word Count'),
+        help=_(
+            'This is the minimum number of words required '
+            'for this question'
+        ),
+        default=0,
+        values={'min': 1},
```
stvstnfrd, line 108: This seems counterintuitive, having a default of 0 but a min of 1;
how does the code handle this validation?

freetextresponse/freetextresponse.py
```diff
@@ -0,0 +1,486 @@
...
+        display_name=_('Weight'),
+        help=_(
+            'This assigns an integer value representing '
+            'the weight of this problem'
+        ),
+        default=1,
+        values={'min': 1},
+        scope=Scope.settings,
+    )
+
```
stvstnfrd, line 130: For consistency, let's remove this newline.

freetextresponse/freetextresponse.py
```diff
@@ -0,0 +1,486 @@
...
+    )
+    score = Float(
+        default=0.0,
+        scope=Scope.user_state,
+    )
+    student_answer = String(
+        default='',
+        scope=Scope.user_state,
+    )
+
```
stvstnfrd, line 143: Remove newlines between fields?

freetextresponse/freetextresponse.py
```diff
@@ -0,0 +1,486 @@
...
+    has_score = True
+
```
stvstnfrd, line 145: Remove newlines between fields?

freetextresponse/freetextresponse.py
```diff
@@ -0,0 +1,486 @@
...
+        Helper method to generate a ValidationMessage from
+        the supplied string
+        """
+        result = ValidationMessage(
+            ValidationMessage.ERROR,
+            unicode(msg)
+        )
+        return result
+
+    def validate_field_data(self, validation, data):
```
stvstnfrd, line 197: Is this used by the mixin?

azimpradhan, line 197: yes

freetextresponse/freetextresponse.py
azimpradhan, line 108: I agree that it may seem counterintuitive, but this is just how the mixin works. The default value can be anything, as it is not subject to validation. However, once the user picks a value, the value picked must be either the default value or a value that meets the constraints above (min &/or max). Additionally, both min and max must be positive. Hope this clarifies your doubts.

freetextresponse/freetextresponse.py
azimpradhan, line 130: Actually the spacing is intentional. I grouped all Scope.settings variables together in alphabetical order (without spaces in between), and did the same for all Scope.user_state variables.

freetextresponse/freetextresponse.py
azimpradhan, line 143: see above

freetextresponse/freetextresponse.py
azimpradhan, line 145: see above

README.markdown
azimpradhan, line 24: since this says e.g. it is intended to serve as an example and not a specific SHA

README.markdown
azimpradhan: yes, I can add that

README.markdown
stvstnfrd, line 24: That's fine. We'll ultimately update it with the package name after we upload to pypi.

freetextresponse/freetextresponse.py
stvstnfrd, line 108: Thanks for the explanation.

freetextresponse/freetextresponse.py
stvstnfrd, line 130: Gotcha.

freetextresponse/freetextresponse.py
stvstnfrd, line 143: +1

freetextresponse/freetextresponse.py
stvstnfrd, line 145: +1

stvstnfrd:
@azimpradhan I left one comment about the name of the installed package, otherwise +1.

After that, merge when ready.

PR #3, opened 2016-01-31
Upgrade to Version 0.1.1

azimpradhan:
Add Submission Received Message to xblock

Change minimum word count from 0 to 1

Create CHANGELOG and update version number


azimpradhan:
@stvstnfrd, @caesar2164, please review.

freetextresponse/freetextresponse.py
```diff
@@ -213,9 +226,9 @@ def validate_field_data(self, validation, data):
                 'Maximum Word Count cannot be negative'
             )
             validation.add(msg)
-        if data.min_word_count < 0:
+        if data.min_word_count < 1:
             msg = FreeTextResponse._generate_validation_message(
-                'Minimum Word Count cannot be negative'
+                'Minimum Word Count cannot be less than 1'
```
stvstnfrd, line 231: So we're now saying that no response can be empty, correct?

freetextresponse/freetextresponse.py
```diff
@@ -223,6 +236,11 @@ def validate_field_data(self, validation, data):
                 'Minimum Word Count cannot be greater than Max Word Count'
             )
             validation.add(msg)
+        if not data.submitted_message:
+            msg = FreeTextResponse._generate_validation_message(
+                'Submission Received Message cannot be blank'
```
stvstnfrd, line 241: Does `_generate_validation_message` handle i18n?

freetextresponse/freetextresponse.py
```diff
@@ -446,18 +474,21 @@ def submit(self, data, suffix=''):
                 )
             )
         self.student_answer = data['student_answer']
-        if self.max_attempts == 0:
-            self.count_attempts = 1
-        else:
-            self.count_attempts += 1
-        self._compute_score()
+        word_count_message = self._get_word_count_message(1)
```
stvstnfrd: Why is this hard-coded?

freetextresponse/freetextresponse.py
```diff
@@ -446,18 +474,21 @@ def submit(self, data, suffix=''):
...
+        if self._word_count_valid():
```
stvstnfrd: What if the count is invalid?

CHANGELOG.md
```diff
@@ -0,0 +1,15 @@
+Changelog
+=========
+Version 0.1.0
+
+				Initial release
```
stvstnfrd: It looks like some hard tabs snuck in here; mind replacing them with 4 space soft tabs?

stvstnfrd:
End of round 1 review.

CHANGELOG.md
azimpradhan: I think hard tabs are required in order for the markdown to properly render. When I replace with spaces, it does not look pretty like this:
https://github.com/Stanford-Online/xblock-free-text-response/blob/azimpradhan/add-srm-freetextresponse/CHANGELOG.md

freetextresponse/freetextresponse.py
azimpradhan: If the count is invalid, then we do not make the student's response count as an attempt, and we do not update the student's score.

freetextresponse/freetextresponse.py
azimpradhan, line 241: looking at the source code, it does not:
https://github.com/edx/XBlock/blob/master/xblock/validation.py
I will i18n the messages in _generate_validation_message

freetextresponse/freetextresponse.py
azimpradhan, line 231: that's correct

CHANGELOG.md
stvstnfrd: In this case, it actually looks like using an unordered list would be more apt, given this is a list of changes.
The version headings can also be converted to markdown headings with a `##` prefix

``` markdown
## Version 0.1.0
- Add Submission Received Message to xblock
- Complete pylint and pep8 compliance
- Change minimum word count from 0 to 1
- Process score only if word count is valid
```

would render

## Version 0.1.0
- Add Submission Received Message to xblock
- Complete pylint and pep8 compliance
- Change minimum word count from 0 to 1
- Process score only if word count is valid

freetextresponse/freetextresponse.py
stvstnfrd: Got it.

CHANGELOG.md
azimpradhan: sounds good, i'll do that

freetextresponse/freetextresponse.py
azimpradhan: This has to do with ignoring the fact the count_attempts might be zero. However, I am going to find a cleaner way of doing this when I commit the changes from this review

freetextresponse/freetextresponse.py
```diff
@@ -279,12 +295,13 @@ def _get_indicator_visiblity_class(self):
...
         return result
 
-    def _get_word_count_message(self):
+    def _get_word_count_message(self, ignore_attempts=False):
         """
         Returns the word count message based on the student's answer
         """
         result = ''
-        if self.count_attempts > 0 and not self._word_count_valid():
+        if (ignore_attempts or self.count_attempts > 0) and \
```
caesar2164: maybe reorganize like so?

``` python
if (
    (ignore_attempts or self.count_attempts > 0)
    and not self._word_count_valid()
):
```

freetextresponse/private/view.html
```diff
@@ -6,6 +6,9 @@ <h2 class="problem-header">{self.display_name}</h2>
         <textarea class="student_answer" style="height:150px" rows="20" cols="70">{self.student_answer}</textarea>
         <span class="status {indicator_visibility_class}" aria-describedby="student_answer"></span>
     </div>
+    <div class="submission-received">
+        {submitted_message}
```
caesar2164, line 10: What does this look like in loco?

'cause from the code it looks like this is rendered as text inside of a `<div>` tag which isn't optimal...

caesar2164:
apart from my two comments, this seems good!

freetextresponse/private/view.html
stvstnfrd, line 10: It looks like we're doing that elsewhere (see line 13).
Just wrap it in a `<span>`?

caesar2164, line 10: You're right, line 13 seems to be sub-optimal as well :wink:

stvstnfrd:
@azimpradhan Have we bumped the version in `edx-platform` for this?

dcadams:
This xblock currently has a sha defining which commit to use in edx-platform. If we start using a version # how can one tell what commit of the code is being used?

azimpradhan:
I think the plan is to keep the defining sha for now. The version number is for when we switch over to PyPi

PR #6, opened 2016-10-06
Change score calculation to be independent of weight

kluo:
@stvstnfrd 

Led to stale progress page data when weight was changed from 0 to non-zero.

### DIFF
--- a/freetextresponse/freetextresponse.py
+++ b/freetextresponse/freetextresponse.py
@@ -374,7 +374,8 @@ class FreeTextResponse(StudioEditableXBlockMixin, XBlock):
                 )
             )
         else:
-            score_string = '{0:g}'.format(self.score)
+            scaled_score = self.score * self.weight
+            score_string = '{0:g}'.format(scaled_score)
             result = "({})".format(
                 ungettext(
                     "{score_string}/{weight} point",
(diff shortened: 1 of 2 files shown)


freetextresponse/freetextresponse.py
```diff
@@ -393,18 +393,13 @@ def _compute_score(self):
...
-        else:
-            self.score = 0.0
+        self.score = credit
         self.runtime.publish(
             self,
             'grade',
             {
                 'value': self.score,
-                'max_value': self.weight
+                'max_value': 1.0
```
stvstnfrd: Should we set this to `Credit.full`?

stvstnfrd:
This matches my understanding of the issue.

@dcadams Have additional thoughts?

stvstnfrd:
I'm wondering if we should even keep `weight` in here at this point; 
what purpose does it serve?

I see we're using it for some of the UI, but will that be misleading if it has to stay in-sync with the grading scale set in CMS? Should we instead just say "50%" or "100%" or "0%"?

We may want to check with CourseOps to see what the intent here.

freetextresponse/freetextresponse.py
```diff
@@ -393,18 +393,13 @@ def _compute_score(self):
...
         """
         credit = self._determine_credit()
-        if credit == Credit.full:
-            self.score = self.weight
-        elif credit == Credit.half:
-            self.score = float(self.weight)/2
-        else:
-            self.score = 0.0
+        self.score = credit
         self.runtime.publish(
```
dcadams, line 398: I think a manifestation of this problem is that "score" in coursware_studentmodule.state has been getting set to weight e.g. '{"student_answer": "What do you expect of me? I have no instructions.", "score": 0, "count_attempts": 1}'
whereas I believe it should be set to the score. This is probably how it is able to now recalculate grade after the weight has been changed. Can you confirm my understanding is correct?

kluo: Right

kluo:
@stvstnfrd  Well you still need weight for when you have multiple problems in a subsection, then the weight matters for how much the free text response should count for that sections overall grade

stvstnfrd:
+1

PR #7, opened 2016-10-09
Save message update story

caseylitton:
@caesar2164 @stvstnfrd PR for updating free text response.   Need to rebase to Kevin's scoring update.
1.  Increase the size of prompt input text box - independent of the following two.
2. Add a save button - this prompted the need for capa update.
3. Update UI to look like capa problems - I'll double check for consistency with Monica next week.
   -. Added a bunch of scenarios that were helpful during testing.

Notes:
-I left in some CSS that should only be used in the workbench for testing.  Do we keep that?
-The textarea css was always overridden in platform.  I could not find a way around needing inline css for coloring the text area border to red/green/default.  The text area previously had in-line css for the textarea height, I just added more.


freetextresponse/freetextresponse.py
```diff
@@ -494,13 +525,44 @@ def submit(self, data, suffix=''):
...
         return result
 
+    @XBlock.json_handler
+    def save_reponse(self, data, suffix=''):
+        # pylint: disable=unused-argument
+        """
+        Processes the user's save
+        """
+        if self.max_attempts > 0 and self.count_attempts >= self.max_attempts:
+            raise StandardError(
```
dcadams: Doesn't this throw a 500? I know we're already doing this in the submit function but this isn't a server error. Can we just use the logger in both places?

freetextresponse/private/view.js
```diff
@@ -4,46 +4,125 @@ function FreeTextResponseView(runtime, element) {
...
     buttonSubmit.on('click', function () {
         buttonSubmit.text('Checking...');
-        runtime.notify('submit', {
-            message: 'Submitting...',
-            state: 'start'
-        });
+        //runtime.notify('submit', {
+        //    message: 'Submitting...',
+        //    state: 'start'
+        //});
```
caesar2164: We should just delete these lines instead of just commenting them...

freetextresponse/private/view.js
```diff
@@ -4,46 +4,125 @@ function FreeTextResponseView(runtime, element) {
...
             }),
             success: function buttonSubmitOnSuccess(response) {
-                textareaParent.removeClass();
-                textareaParent.addClass(userInputClass);
-                textareaParent.addClass(response.indicator_class);
                 usedAttemptsFeedback.text(response.used_attempts_feedback);
-                submitParent.removeClass();
-                submitParent.addClass(response.submit_class);
+                //submitParent.removeClass();
+                //submitParent.addClass(response.submit_class);
```
caesar2164: remove commented lines

freetextresponse/private/view.js
```diff
@@ -4,46 +4,125 @@ function FreeTextResponseView(runtime, element) {
...
                 problemProgress.text(response.problem_progress);
-                wordCountError.text(response.word_count_message);
                 submissionReceivedMessage.text(response.submitted_message);
                 buttonSubmit.text('Submit');
+                userAlertMessage.text(response.user_alert);
+                //saveParent.removeClass();
+                //saveParent.addClass(response.submit_class);
```
caesar2164: remove comments

freetextresponse/private/view.js
```diff
@@ -4,46 +4,125 @@ function FreeTextResponseView(runtime, element) {
...
+                
+                // Update Indicators
+                _setIndicators(response.indicator_class, response.indicator_visibility_class);
 
-                runtime.notify('submit', {
-                    state: 'end'
-                });
+                //runtime.notify('submit', {
+                //    state: 'end'
+                //});
```
caesar2164: remove commented code

freetextresponse/private/view.js
```diff
@@ -4,46 +4,125 @@ function FreeTextResponseView(runtime, element) {
...
         });
         return false;
     });
+
+    buttonSave.on('click', function () {
+        buttonSave.text('Checking...');
+        //runtime.notify('save', {
+        //    message: 'Saving...',
+        //    state: 'start'
+        //});
```
caesar2164: remove commented code

freetextresponse/private/view.js
```diff
@@ -4,46 +4,125 @@ function FreeTextResponseView(runtime, element) {
...
+                submissionReceivedMessage.text(response.submitted_message);
+                buttonSave.text('Save');
+                userAlertMessage.text(response.user_alert);
+                
+                // Update Indicators
+                _setIndicators(response.indicator_class, response.indicator_visibility_class);
+                
+                //runtime.notify('save', {
+                //    state: 'end'
+                //});
```
caesar2164: remove commented code

freetextresponse/public/view.js
```diff
@@ -4,46 +4,125 @@ function FreeTextResponseView(runtime, element) {
...
     buttonSubmit.on('click', function () {
         buttonSubmit.text('Checking...');
-        runtime.notify('submit', {
-            message: 'Submitting...',
-            state: 'start'
-        });
+        //runtime.notify('submit', {
+        //    message: 'Submitting...',
+        //    state: 'start'
+        //});
```
caesar2164: remove commented code

freetextresponse/public/view.js
```diff
@@ -4,46 +4,125 @@ function FreeTextResponseView(runtime, element) {
...
             }),
             success: function buttonSubmitOnSuccess(response) {
-                textareaParent.removeClass();
-                textareaParent.addClass(userInputClass);
-                textareaParent.addClass(response.indicator_class);
                 usedAttemptsFeedback.text(response.used_attempts_feedback);
-                submitParent.removeClass();
-                submitParent.addClass(response.submit_class);
+                //submitParent.removeClass();
+                //submitParent.addClass(response.submit_class);
```
caesar2164: remove commented code

freetextresponse/public/view.js
```diff
@@ -4,46 +4,125 @@ function FreeTextResponseView(runtime, element) {
...
                 problemProgress.text(response.problem_progress);
-                wordCountError.text(response.word_count_message);
                 submissionReceivedMessage.text(response.submitted_message);
                 buttonSubmit.text('Submit');
+                userAlertMessage.text(response.user_alert);
+                //saveParent.removeClass();
+                //saveParent.addClass(response.submit_class);
```
caesar2164: remove commented code

freetextresponse/public/view.js
```diff
@@ -4,46 +4,125 @@ function FreeTextResponseView(runtime, element) {
...
+                
+                // Update Indicators
+                _setIndicators(response.indicator_class, response.indicator_visibility_class);
 
-                runtime.notify('submit', {
-                    state: 'end'
-                });
+                //runtime.notify('submit', {
+                //    state: 'end'
+                //});
```
caesar2164: remove commented code

freetextresponse/public/view.js
```diff
@@ -4,46 +4,125 @@ function FreeTextResponseView(runtime, element) {
...
         });
         return false;
     });
+
+    buttonSave.on('click', function () {
+        buttonSave.text('Checking...');
+        //runtime.notify('save', {
+        //    message: 'Saving...',
+        //    state: 'start'
+        //});
```
caesar2164: remove commented code

freetextresponse/public/view.js
```diff
@@ -4,46 +4,125 @@ function FreeTextResponseView(runtime, element) {
...
+                submissionReceivedMessage.text(response.submitted_message);
+                buttonSave.text('Save');
+                userAlertMessage.text(response.user_alert);
+                
+                // Update Indicators
+                _setIndicators(response.indicator_class, response.indicator_visibility_class);
+                
+                //runtime.notify('save', {
+                //    state: 'end'
+                //});
```
caesar2164: remove commented code

caesar2164 (changes requested):

freetextresponse/freetextresponse.py
```diff
@@ -21,7 +21,7 @@
 
 
 class FreeTextResponse(StudioEditableXBlockMixin, XBlock):
-    #  pylint: disable=too-many-ancestors, too-many-instance-attributes
+    #  pylint: disable=too-many-ancestors, too/many-instance-attributes
```
stvstnfrd: I don't think you meant to do this?

freetextresponse/freetextresponse.py
```diff
@@ -33,8 +33,13 @@ def workbench_scenarios():
         scenarios = [
             ('Free-text Response XBlock',
              '''<sequence_demo>
-                    <freetextresponse />
```
stvstnfrd, line 38: I would recommend keeping this one (with no args) so that you are still testing the default case.

freetextresponse/freetextresponse.py
```diff
@@ -33,8 +33,13 @@ def workbench_scenarios():
...
-                    <freetextresponse name='My First XBlock' />
+                    <freetextresponse display_name='Full Credit is asdf, half is fdsa' fullcredit_keyphrases="['asdf']" halfcredit_keyphrases="['fdsa']" />
+                    <freetextresponse display_name='Default Free-text Response'/>
```
stvstnfrd: Since this is XML, these should probably be double quotes (I know the original one here wasn't).

stvstnfrd: This applies to each scenario here..

freetextresponse/freetextresponse.py
```diff
@@ -136,6 +142,15 @@ def workbench_scenarios():
         values={'min': 1},
         scope=Scope.settings,
     )
+    saved_message = String(
+        display_name=_('Draft Received Message'),
+        help=_(
+            'This is the message students will see upon '
+            'submitting a draft response'
+        ),
+        default=_("Your answers have been saved but not graded. Click 'Submit' to grade them."),
```
stvstnfrd: Not only do we prefer single quotes for non-interpolated strings, but doing so here on the outer quotes will allow us to use double quotes around `Submit` (without escaping), which is probably a bit more correct grammatically.

freetextresponse/freetextresponse.py
```diff
@@ -149,6 +164,10 @@ def workbench_scenarios():
         default='',
         scope=Scope.user_state,
     )
+    student_answer_type = String(
```
stvstnfrd: I don't care for this;
it looks like a concurrency/state management issue waiting to happen.

It looks like the motivation for this is that you reference this value
w/in some of your helper methods?
Instead of making it a preserved, instance variable,
we may want to just explicitly pass the value into the helpers, like:

```
function _get_submitted_message(answer_type) {
    if ('save' === answer_type) {
        // blah
    }
}
```

instead of 

```
function _get_submitted_message() {
    if ('save' === self.answer_type) {
        // blah
    }
}
```

Otherwise you risk a number of problems, everything from cached results to issues from simultaneous requests.

(BTW: I realized I switched between PY/JS syntax above; hopefully it still makes sense)

freetextresponse/freetextresponse.py
stvstnfrd: Yeah, we definitely shouldn't be throwing an error here.
I'm not even convinced (though I could be) we have to log this; do we otherwise log when people reach max attempts?

freetextresponse/private/view.js
```diff
@@ -4,46 +4,125 @@ function FreeTextResponseView(runtime, element) {
     var $ = window.jQuery;
     var $element = $(element);
     var buttonSubmit = $element.find('.check.Submit');
-    var textareaParent = $element.find('.student_answer').parent();
+    var textareaParent = $element.find('.textarea').parent();
```
stvstnfrd: Ewww, naming a class the same thing as the actual element type is poor form.
Do you even need the class? If there's only one `textarea` per XBlock, this can just be:

``` javascript
$element.find('TEXTAREA')
```

If there could be multiple, chose a more descriptive classname, though I suspect it's superfluous and we can just reference the element type.

freetextresponse/private/view.js
```diff
@@ -4,46 +4,125 @@ function FreeTextResponseView(runtime, element) {
...
-    var wordCountError = $element.find('.word-count-error');
     var submissionReceivedMessage = $element.find('.submission-received');
     var url = runtime.handlerUrl(element, 'submit');
     var userInputClass = 'user-input';
+    var buttonSave = $element.find('.save');
+    var saveParent = $element.find('.save').parent();
+    var urlSave = runtime.handlerUrl(element, 'save_reponse');
+    var userAlertMessage = $element.find('.user_alert');
+    var statusIndicator = $element.find('.status');
+    var textarea = $element.find('.textarea');
```
stvstnfrd: This could probably afford to have a more descriptive name, something that describes what it's for, not just what it is, `studentAnswer` or some such..

stvstnfrd: I would move this above line 7 so you end up with something of the form:

``` javascript
var answer = $element.find('TEXTAREA');
var answerParent = answer.parent();
var answerGrandParent = answerParent.parent();
```

This will save on superfluous `find`s.

freetextresponse/private/view.js
```diff
@@ -4,46 +4,125 @@ function FreeTextResponseView(runtime, element) {
...
+    var textareaParentParent = textareaParent.parent();
```
stvstnfrd: This is a terribly opaque name. What is the grandparent for? Is there a more descriptive name we can use?

freetextresponse/private/view.js
```diff
@@ -4,46 +4,125 @@ function FreeTextResponseView(runtime, element) {
...
+    
+    function _setBorderColor(indicator_class) {
+        if ( 'correct' == indicator_class ) {
```
stvstnfrd: Unless you have explicit reasons to the contrary, always use strict equality checking in JavaScript with `===` over `==`.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness

freetextresponse/private/view.js
```diff
@@ -4,46 +4,125 @@ function FreeTextResponseView(runtime, element) {
...
+            textarea.css('border', 'solid #1e9348');
+        } else if ( 'incorrect' == indicator_class ) {
+            textarea.css('border', 'solid #b20610');
+        } else {
+            textarea.css('border', 'solid #e4e4e4');
```
stvstnfrd: This is a bit backward.

If we already have a class, we should just add/remove it to the element;
JavaScript should have no knowledge of the CSS styling rules.
This is more difficult to maintain.

freetextresponse/private/view.js
```diff
@@ -4,46 +4,125 @@ function FreeTextResponseView(runtime, element) {
...
+        }
+    }
+
+    function _setIndicators(indicator_class, indicator_visibility_class) {
+        // Set Indicator icon
+        textareaParent.removeClass('correct');
```
stvstnfrd, line 24: We shouldn't need to add/remove this class to/from multiple components;
we should leverage the Cascading nature of stylesheets.

I would just add the class directly to the top-level `$element` and then reference children in CSS like:

``` javascript
DIV.my_xblock.correct DIV.capa_inputtype {
    /* RULES FOR GRAND_PARENT */
}
DIV.my_xblock.correct DIV.capa_inputtype DIV {
    /* RULES FOR PARENT */
}
```

freetextresponse/private/view.js
```diff
@@ -4,46 +4,125 @@ function FreeTextResponseView(runtime, element) {
...
+        textareaParent.removeClass('incorrect');
+        textareaParent.removeClass('unanswered');
+        textareaParent.addClass(indicator_class);
+        textareaParentParent.removeClass('correct');
+        textareaParentParent.removeClass('incorrect');
+        textareaParentParent.removeClass('unanswered');
+        textareaParentParent.addClass(indicator_class);
+        //Border Color
```
stvstnfrd: Superfluous. `//Border Color` is even less informative than `_setBorderColor` ;)

freetextresponse/private/view.js
```diff
@@ -4,46 +4,125 @@ function FreeTextResponseView(runtime, element) {
...
+        if ( 'hidden' == indicator_visibility_class ) {
+            _setBorderColor('unanswered');
```
stvstnfrd: Same idea as above; JS shouldn't do the heavy lifting here.
It should just mark the entire XBlock as either correct or incorrect
`$element.addClass('whatever')` and let the CSS handle any styling.

freetextresponse/private/view.js
stvstnfrd: @caesar2164 I think these are debug-related and _shouldn't_ be deleted, right @caseylitton ?

freetextresponse/private/view.js
```diff
@@ -4,46 +4,125 @@ function FreeTextResponseView(runtime, element) {
...
+    buttonSave.on('click', function () {
+        buttonSave.text('Checking...');
+        //runtime.notify('save', {
+        //    message: 'Saving...',
+        //    state: 'start'
+        //});
+        $.ajax(urlSave, {
+            type: 'POST',
+            data: JSON.stringify({
+                'student_answer': $element.find('.student_answer').val()
```
stvstnfrd, line 71: Haven't we already looked up the `student_answer TEXTAREA` up above?
I don't think we need another lookup here..

freetextresponse/private/view.less
```diff
@@ -1,6 +1,8 @@
 .freetextresponse {
 
-  .user-input {
+  //  DEBUG - should have access to built-int
```
stvstnfrd: `s/built-int/built-in`?

freetextresponse/private/view.less
```diff
@@ -66,11 +74,17 @@
...
+    border: 1px solid #ebe8bf;
+    border-radius: 3px;
+    background: #fffcdd;
+    font-size: 0.9em;
   }
 
-  .submission-received {
-    color: #2EA41E;
-  }
+  // Found on internet
```
stvstnfrd: ??

freetextresponse/private/view.less
```diff
@@ -66,11 +74,17 @@
     display: inline-block;
   }
 
-  .word-count-error {
-    color: #A50000;
+  //  DEBUG - should have access to built-int
+  //  capa_alert in need in lms and studio
```
stvstnfrd: What's the reference, that these should be in LMS but are missing from SDK?

freetextresponse/freetextresponse.py
```diff
@@ -178,10 +197,11 @@ def student_view(self, context=None):
             used_attempts_feedback=self._get_used_attempts_feedback(),
             submit_class=self._get_submit_class(),
             indicator_visibility_class=self._get_indicator_visiblity_class(),
-            word_count_message=self._get_word_count_message(
-                self.count_attempts
-            ),
             submitted_message=self._get_submitted_message(),
+            save_class=self._get_submit_class(),
```
stvstnfrd: Why are `submit_class` and `save_class` each set to `_get_submit_class`?

freetextresponse/private/view.html
```diff
@@ -1,24 +1,29 @@
...
-    <div class="submission-received">
-        {submitted_message}
-    </div>
-    <div class="word-count-error">
-        {word_count_message}
-    </div>
-    <div class="action">
-        <span class="{submit_class}">
-            <button class="check Submit" data-checking="Checking..." data-value="Submit" aria-disabled="false">
+    <div class="problem">
```
stvstnfrd: Do we really need this extra div?
Could we get away with adding `class=problem` to the top-level `freetextresponse` div above?

freetextresponse/private/view.html
```diff
@@ -1,24 +1,29 @@
...
+        <p>{self.prompt}</p>
+        <div class="capa_inputtype textline {indicator_class}">
+            <div class="{indicator_class}">
+                <textarea class="textarea student_answer" style="height:150px; box-sizing:border-box; border-radius:3px; border:2px solid #e4e4e4; min-width:160px;" rows="20" cols="70">{self.student_answer}</textarea>
```
stvstnfrd: I know this predates you, but we should move this to SASS, instead of styling inline.

freetextresponse/private/view.html
```diff
@@ -1,24 +1,29 @@
...
+                <span class="status {indicator_visibility_class}" aria-describedby="student_answer"></span>
+            </div>
+        </div>
+        <div class="capa_alert submission-received">
+            {submitted_message}
+        </div>
+        <div class="action">
+            <button class="check Submit" data-checking="Checking..." data-value="Submit">
```
stvstnfrd: @caesar2164 What is preferred way to do this again, `&hellip;`?

freetextresponse/private/view.js
```diff
@@ -4,46 +4,125 @@ function FreeTextResponseView(runtime, element) {
...
     var usedAttemptsFeedback = $element.find('.action .used-attempts-feedback');
     var problemProgress = $element.find('.problem-progress');
     var submitParent = $element.find('.Submit').parent();
-    var wordCountError = $element.find('.word-count-error');
     var submissionReceivedMessage = $element.find('.submission-received');
     var url = runtime.handlerUrl(element, 'submit');
     var userInputClass = 'user-input';
+    var buttonSave = $element.find('.save');
+    var saveParent = $element.find('.save').parent();
+    var urlSave = runtime.handlerUrl(element, 'save_reponse');
```
stvstnfrd, line 16: Maybe make this stand out more by moving it to either the top or bottom of this stanza, instead of nesting it into the middle of CSS selectors (since it's an ajax handler).

freetextresponse/private/view.js
```diff
@@ -4,46 +4,125 @@ function FreeTextResponseView(runtime, element) {
...
+                //});
+            },
+            error: function buttonSaveOnError() {
+                runtime.notify('error', {});
+            }
+        });
+        return false;
+    });
+
+    textareaParent.on('keydown', function() {
```
stvstnfrd: Nit: Always put a space after `function` (you did this above).
Also on the consistency note, any reason you're not naming this anonymous function,
whereas you are doing so above?

stvstnfrd (changes requested):
First pass complete.

freetextresponse/private/view.js
caseylitton: @caesar2164 yes, I left these commented out for testing and an FYI that 'notify' does not exist in 'runtime' on workbench.

freetextresponse/freetextresponse.py
caseylitton: Fixed

freetextresponse/freetextresponse.py
caseylitton: The original functionality, which may not exist in capa problems, is to hide the submit button with a 'nodisplay....something' tag if the user reached max attempts.  In any case, I figured if no submission then no saving sounded right.

Edit: the functionality does exist in capa problems, when max attempts is reached the submit/save buttons are hidden

freetextresponse/private/view.js
caseylitton: ahh...yea, I meant to go back and change the textarea and textareaParentParent names.  The original html nesting changed a bit at the end to match with capa

freetextresponse/private/view.js
caseylitton: The textarea css had to be inline html?   That was the only way I could get the border to change color.  I'll clean up everything else and we can look at this in more detail.

freetextresponse/private/view.js
caseylitton: I'm not following, this is consistent with the button click event callbacks.

freetextresponse/private/view.less
caseylitton: They don't exist in the workbench but do exist in LMS.  So to view the functionality in workbench I kept the css classes.  "should have access to capa_alert class in lms and studio"

freetextresponse/private/view.less
caseylitton: old css thing I was using before switching to capa css classes.  Left in the comment.

freetextresponse/private/view.html
caseylitton: This is modeled after the capa problems, adding it to freetextresponse would probably be fine though.

freetextresponse/private/view.html
caseylitton: I could not find a way around the inline style for the textarea.  It is original code(minus the border).  The height was always set inline.    Seemed like devstack kept overriding the textarea css.

freetextresponse/freetextresponse.py
caseylitton: Removed.

freetextresponse/freetextresponse.py
caseylitton: Fixed.  Changed to nodisplay_class and used for both buttons.

freetextresponse/freetextresponse.py
caseylitton: Fixed

freetextresponse/freetextresponse.py
caseylitton: Sure, I thought about doing that way but nothing else was using that method.  Fixed

freetextresponse/private/view.js
caseylitton: Fixed

freetextresponse/private/view.js
caseylitton: Fixed

freetextresponse/private/view.js
caseylitton: Fixed

caseylitton: I wanted to clean up the variable declaration too, it wasn't clean to begin with.  I'll do a clean up toward the end.  There is a mis-spelled word in the freetextresponse.py that I wanted to fix too.

freetextresponse/private/view.js
caseylitton: fixed, textareaParent was not needed. just used grandparent and textareaStudentAnswer

freetextresponse/private/view.js
caseylitton: Fixed,  I know this just forgot from using python

freetextresponse/private/view.js
caseylitton: yea, we didn't need the textareaParent

freetextresponse/private/view.js
stvstnfrd: @caseylitton FYI: Rather than commenting that out multiple times throughout the codebase,
it may be better served to polyfill the method. You could do something like:

``` javascript
// The "or" (`||`) makes sure we don't redefine the method if it already exists.
runtime.notify = runtime.notify || function () {
    // The logging statement is optional.
    console.log('POLYFILL runtime.notify', arguments);
};
```

This way, you only make one additive change, instead of removing code in many places.
The latter can be prone to mistakes.

freetextresponse/freetextresponse.py
```diff
@@ -318,17 +334,16 @@ def _get_word_count_message(self, ignore_attempts=False):
             )
         return result
 
-    def _get_indicator_class(self):
+    def _get_indicator_class(self, student_answer_type=None):
```
stvstnfrd: It looks like we always pass in an arg for `student_answer_type`,
so we shouldn't default to `None`; we always want it to be one of two strings.

``` python
def _get_indicator_class(self, student_answer_type):
    pass
```

caseylitton: student_answer_type=None is an important part of the code.  It's how we get the 'unanswered' indicator.  There are three strings, 'unanswered', 'correct', and 'incorrect'

freetextresponse/private/view.js
caseylitton: nice

caseylitton:
@caesar2164 @stvstnfrd PR for capa UI updates for free text response.

freetextresponse/freetextresponse.py
```diff
@@ -459,45 +500,81 @@ def _get_submit_class(self):
...
         return result
 
-    def _get_submitted_message(self):
+    def _get_submitted_message(self, submit_type=None):
         """
         Returns the message to display in the submission-received div
         """
         result = ''
-        if self.count_attempts > 0 and self._word_count_valid():
+        if 'submit' == submit_type and self._word_count_valid():
```
caesar2164: logically this would make more sense:
`if submit_type == 'submit' and self._word_count_valid():`

freetextresponse/freetextresponse.py
```diff
@@ -459,45 +500,81 @@ def _get_submit_class(self):
...
             result = self.submitted_message
         return result
 
+    def _get_user_alert(self, submit_type=None, ignore_attempts=False):
+        """
+        Returns the message to display in the user_alert(TBD) div
+        """
+        result = ''
+        if 'submit' == submit_type and not self._word_count_valid():
```
caesar2164: Same as line 508 above. Also is there any way to not repeat code?

freetextresponse/private/view.less
```diff
@@ -24,12 +24,40 @@
...
+    &.correct .student_answer {
+      border:2px solid #1e9348;
+    }
+    
+    &.incorrect .student_answer {
+      border:2px solid #b20610;
+    }
+    
   }
+
```
caesar2164: remove extra line

freetextresponse/public/view.js
```diff
@@ -4,46 +4,109 @@ function FreeTextResponseView(runtime, element) {
...
+        console.log('POLYFILL runtime.notify', arguments);
+    };
 
     buttonSubmit.on('click', function () {
         buttonSubmit.text('Checking...');
         runtime.notify('submit', {
             message: 'Submitting...',
             state: 'start'
         });
+
```
caesar2164: remove

freetextresponse/freetextresponse.py
```diff
@@ -459,45 +500,81 @@ def _get_submit_class(self):
...
+    @XBlock.json_handler
+    def save_reponse(self, data, suffix=''):
+        # pylint: disable=unused-argument
+        """
+        Processes the user's save
+        """
+        if ((0 == self.max_attempts) or
+                (0 < self.max_attempts and
+                    self.count_attempts < self.max_attempts)):
+            self.student_answer = data['student_answer']
```
caesar2164, line 571: this conditional seems unnecessary.

caesar2164:
This looks pretty good, but there's some convoluted code that should probably be cleaned up while it's being messed with.

caseylitton:
@stvstnfrd @caesar2164 I'm going to make a clean up commit since both of you have expressed interest in cleaning up the old code .  TBDone

freetextresponse/freetextresponse.py
stvstnfrd: @caesar2164 I don't think "logically" is the proper term here.

There is a good argument to be made that, for some languages, placing the literal primitive on the LHS is preferred. Specifically, this reduces a class of bugs where you accidentally omit an `=` sign and mistakenly perform an assignment, which then may flip the expression from false to truthy.

``` javascript
var x = 10;
if (x = 20) {
    console.log('true');
} else {
    console.log('false');
}
console.log(x);
```

This will print `true` and `20`, whereas you almost surely meant `false` and `10`.

That all said, I'm inclined to avoid this style for 2 reasons:
- In python, assignment is a statement not an expression, so the above mistake is syntactically invalid anyway.
- We don't follow this convention anywhere else in the codebase.

freetextresponse/private/view.js
```diff
@@ -4,46 +4,109 @@ function FreeTextResponseView(runtime, element) {
...
+    });
+
+    textareaStudentAnswer.on('keydown', function() {
+        // Reset Messages
+        submissionReceivedMessage.text('');
+        userAlertMessage.text('');
+        textareaParent.removeClass('correct');
+        textareaParent.removeClass('incorrect');
+        textareaParent.removeClass('unanswered');
+        textareaParent.addClass('unanswered');
```
stvstnfrd: It looks like you're repeating these 4 lines a couple of times. Move into a helper method:

``` javascript
function setClassForTextAreaParent(new_class) {
    // remove old ones
    // add new one
}
```

freetextresponse/freetextresponse.py
stvstnfrd: Yeah, you could do something like:

``` python
if max_attempts <= 0 or count_attempts < max_attempts:
    pass
```

freetextresponse/freetextresponse.py
```diff
@@ -459,45 +500,81 @@ def _get_submit_class(self):
...
-                self.count_attempts += 1
-            self._compute_score()
+        if ((0 == self.max_attempts) or
+                (0 < self.max_attempts and
+                    self.count_attempts < self.max_attempts)):
+            self.student_answer = data['student_answer']
+            submit_type = 'submit'
+            if self._word_count_valid():
+                if self.max_attempts == 0:
+                    self.count_attempts = 1
```
stvstnfrd: I know this was already here, so just thinking out loud:
Why aren't we always counting attempts?
If `max_attempts` is unlimited (0), shouldn't we still keep count?
Why hard-code this to 1?

freetextresponse/freetextresponse.py
```diff
@@ -459,45 +500,81 @@ def _get_submit_class(self):
...
-            )
-        self.student_answer = data['student_answer']
-        if self._word_count_valid():
-            if self.max_attempts == 0:
-                self.count_attempts = 1
-            else:
-                self.count_attempts += 1
-            self._compute_score()
+        if ((0 == self.max_attempts) or
+                (0 < self.max_attempts and
```
stvstnfrd: See other comment below;
I suspect this can be simpler.

freetextresponse/freetextresponse.py
```diff
@@ -459,45 +500,81 @@ def _get_submit_class(self):
...
-        if self.count_attempts > 0 and self._word_count_valid():
+        if 'submit' == submit_type and self._word_count_valid():
             result = self.submitted_message
         return result
 
+    def _get_user_alert(self, submit_type=None, ignore_attempts=False):
+        """
+        Returns the message to display in the user_alert(TBD) div
+        """
+        result = ''
```
stvstnfrd, line 529: So we don't show a message if we're submitting and the word count is valid?

freetextresponse/private/view.js
```diff
@@ -4,46 +4,109 @@ function FreeTextResponseView(runtime, element) {
...
+    var buttonSave = $element.find('.save');
+    var userAlertMessage = $element.find('.user_alert');
+    var textareaStudentAnswer = $element.find('.student_answer');
+    var textareaParent = textareaStudentAnswer.parent();
+    var url = runtime.handlerUrl(element, 'submit');
+    var urlSave = runtime.handlerUrl(element, 'save_reponse');
+   
+
+    // Override notify if it does not exist. Like in the xblock workbench.
+    // The "or" (`||`) makes sure we don't redefine the method if it already exists.
```
stvstnfrd: We don't need this comment in the code;
I wrote that as a review comment, in case you were unfamiliar
with that idiom in this context.

freetextresponse/private/view.js
```diff
@@ -4,46 +4,109 @@ function FreeTextResponseView(runtime, element) {
...
     var userInputClass = 'user-input';
+    var buttonSave = $element.find('.save');
+    var userAlertMessage = $element.find('.user_alert');
+    var textareaStudentAnswer = $element.find('.student_answer');
+    var textareaParent = textareaStudentAnswer.parent();
+    var url = runtime.handlerUrl(element, 'submit');
+    var urlSave = runtime.handlerUrl(element, 'save_reponse');
+   
+
+    // Override notify if it does not exist. Like in the xblock workbench.
```
stvstnfrd: s/override/polyfill/

freetextresponse/private/view.js
```diff
@@ -4,46 +4,109 @@ function FreeTextResponseView(runtime, element) {
...
+    // The "or" (`||`) makes sure we don't redefine the method if it already exists.
+    runtime.notify = runtime.notify || function () {
+        // The logging statement is optional.
```
stvstnfrd: Again, these were overly annotated examples for the purposes of explaining and justifying my review.

freetextresponse/private/view.js
```diff
@@ -4,46 +4,109 @@ function FreeTextResponseView(runtime, element) {
...
+        console.log('POLYFILL runtime.notify', arguments);
+    };
 
     buttonSubmit.on('click', function () {
         buttonSubmit.text('Checking...');
         runtime.notify('submit', {
             message: 'Submitting...',
             state: 'start'
         });
+
```
stvstnfrd: Drop unrelated change.

freetextresponse/private/view.js
```diff
@@ -4,46 +4,109 @@ function FreeTextResponseView(runtime, element) {
...
+                    textareaParent.removeClass('unanswered');
+                    textareaParent.addClass(response.indicator_class); 
+                }
+                
                 runtime.notify('submit', {
                     state: 'end'
                 });
             },
-            error: function buttonSubmitOnError() {
+            error: function buttonSubmitOnError(err) {
```
stvstnfrd: As this is optional and unrelated, let's drop it from the diff for now.

freetextresponse/private/view.js
```diff
@@ -4,46 +4,109 @@ function FreeTextResponseView(runtime, element) {
...
+                });
+            },
+            error: function buttonSaveOnError() {
+                runtime.notify('error', {});
+            }
+        });
+        return false;
+    });
+
+    textareaStudentAnswer.on('keydown', function() {
```
stvstnfrd, line 93: Elsewhere, you have a space after the `function` keyword, let's do the same here for consistency.

stvstnfrd (changes requested):

freetextresponse/freetextresponse.py
caseylitton: I assumed the limit was in place so it wouldn't have the chance of becoming arbitrarily large.  Prior to this update, there was an assumption that the count will not be greater than max attempts.  I believe this is fixed per the other max_attempts comments

freetextresponse/freetextresponse.py
caseylitton: I made this fix in the updated version.

freetextresponse/freetextresponse.py
caseylitton: We do show it, maybe there was a mistake when you viewed it?

freetextresponse/freetextresponse.py
caseylitton: fixed according to previous max_attempts comments

freetextresponse/private/view.js
caseylitton: I figured, but left it in b/c other people may be unfamiliar.    It's out now

caseylitton:
@stvstnfrd  @caesar2164 This PR should address the comments and some other errors with UI response. Visibility was miss spelled in the freetestresponse.py too.    If this looks good then I can go through it with Monica next week to make sure it looks right in studio/lms

freetextresponse/freetextresponse.py
```diff
@@ -459,45 +498,71 @@ def _get_submit_class(self):
...
         return result
 
-    def _get_submitted_message(self):
+    def _get_submitted_message(self, submit_type=None):
         """
         Returns the message to display in the submission-received div
         """
         result = ''
-        if self.count_attempts > 0 and self._word_count_valid():
+        if self._word_count_valid():
```
caesar2164, line 520: NICE

freetextresponse/freetextresponse.py
```diff
@@ -151,7 +193,7 @@ def workbench_scenarios():
     )
 
     has_score = True
-
+visibility
```
caesar2164: ?

freetextresponse/freetextresponse.py
```diff
@@ -464,40 +503,64 @@ def _get_submitted_message(self):
...
-                )
-            )
-        self.student_answer = data['student_answer']
-        if self._word_count_valid():
-            if self.max_attempts == 0:
-                self.count_attempts = 1
-            else:
-                self.count_attempts += 1
-            self._compute_score()
+        if self.max_attempts == 0 or self.count_attempts < self.max_attempts:
```
caesar2164, line 542: you could also do this:

``` python
unlimited_attempts = self.max_attempts == 0
has_remaining_attempts = self.count_attempts < self.max_attempts

if unlimited_attempts or has_remaining_attempts:
```

caesar2164: which makes it REALLY EASY to read :)

freetextresponse/private/view.html
caesar2164: yeah, this HTML uses pre-existing styling in platform

freetextresponse/private/view.js
```diff
@@ -4,38 +4,52 @@ function FreeTextResponseView(runtime, element) {
...
+
+    // POLYFILL notify if it does not exist. Like in the xblock workbench.
+    runtime.notify = runtime.notify || function () {
+        console.log('POLYFILL runtime.notify', arguments);
+    };
+ 
+    function setClassForTextAreaParent(new_class) {
+        textareaParent.removeClass('correct');
+        textareaParent.removeClass('incorrect');
+        textareaParent.removeClass('unanswered');
```
caesar2164:
```js
textareaParent.removeClass('correct incorrect unanswered');
```
http://api.jquery.com/removeClass/

freetextresponse/private/view.js
```diff
@@ -4,38 +4,52 @@ function FreeTextResponseView(runtime, element) {
...
+        textareaParent.addClass(new_class);
```
caesar2164, line 27:
``` js
textareaParent.removeClass('correct incorrect unanswered');
textareaParent.addClass(new_class);
```

http://api.jquery.com/removeClass/

freetextresponse/private/view.html
caseylitton: I'll check this when doing a final test on lms/studio.

freetextresponse/private/view.js
caesar2164: nvmd

freetextresponse/private/view.js
caesar2164: agreed

freetextresponse/public/view.js
```diff
@@ -4,38 +4,52 @@ function FreeTextResponseView(runtime, element) {
...
     var submissionReceivedMessage = $element.find('.submission-received');
+    var userAlertMessage = $element.find('.user_alert');
+    var textareaStudentAnswer = $element.find('.student_answer');
+    var textareaParent = textareaStudentAnswer.parent();
+
     var url = runtime.handlerUrl(element, 'submit');
-    var userInputClass = 'user-input';
+    var urlSave = runtime.handlerUrl(element, 'save_reponse');
+   
+
```
caesar2164, line 17: remove extra line

caesar2164:
I like the cleanup! this is very close to being ready!

freetextresponse/private/view.html
caseylitton: okay, changed

freetextresponse/freetextresponse.py
caseylitton: @caesar2164 @stvstnfrd 
I'm fine with this.  My only concern is the above conditional is in three different functions.  Should we add it to all three or just submit and save events?  We could add this to save, submit and the default view functions and then pass them into any functions that require them.  Both parts of the conditional are used many times from save, submit, and default

freetextresponse/freetextresponse.py
```diff
@@ -287,7 +327,7 @@ def build_fragment(
             fragment.initialize_js(fragment_js)
         return fragment
 
-    def _get_indicator_visiblity_class(self):
+    def _get_indicator_visibility_class(self):
```
stvstnfrd, line 332: Nice catch.

freetextresponse/freetextresponse.py
```diff
@@ -322,13 +362,12 @@ def _get_indicator_class(self):
...
         """
-        result = ''
-        if self.count_attempts == 0:
-            result = 'unanswered'
-        elif self._determine_credit() == Credit.zero:
-            result = 'incorrect'
-        else:
-            result = 'correct'
+        result = 'unanswered'
+        if self.display_correctness and self._word_count_valid():
```
stvstnfrd, line 380: Doesn't this mean that if the word count is invalid, then it will display `unanswered`?
That doesn't seem correct..

freetextresponse/freetextresponse.py
```diff
@@ -464,40 +503,64 @@ def _get_submitted_message(self):
...
         """
         result = ''
-        if self.count_attempts > 0 and self._word_count_valid():
+        if self._word_count_valid():
             result = self.submitted_message
         return result
 
+    def _get_user_alert(self, ignore_attempts=False):
+        """
+        Returns the message to display in the user_alert(TBD) div
```
stvstnfrd: What's TBD?

freetextresponse/freetextresponse.py
stvstnfrd: Could you move this logic into a helper, like:

``` python
def can_keep_attempting(self):
    unlimited_attempts = self.max_attempts == 0
    has_remaining_attempts = self.count_attempts < self.max_attempts
    return unlimited_attempts or has_remaining_attempts
```

freetextresponse/freetextresponse.py
```diff
@@ -464,40 +503,64 @@ def _get_submitted_message(self):
...
-            if self.max_attempts == 0:
-                self.count_attempts = 1
-            else:
-                self.count_attempts += 1
-            self._compute_score()
+        if self.max_attempts == 0 or self.count_attempts < self.max_attempts:
+            self.student_answer = data['student_answer']
+            if self._word_count_valid():
+                if self.max_attempts == 0:
+                    self.count_attempts = 1
```
stvstnfrd: > I assumed the limit was in place so it wouldn't have the chance of becoming arbitrarily large.

It's an integer; it's going to take up the same amount of space whether the value is one or a million.
Unless, of course, you're worried about integer overflow, in the event that someone makes `9223372036854775807` attempts ;)

stvstnfrd (changes requested):

freetextresponse/freetextresponse.py
caseylitton: @stvstnfrd The UI is 'unanswered' prior to all submissions as a feature.  Except when the user clicked submit twice in a row without changing their answer.  In that case the UI will already be displaying the correctness from the first click.   After submission, the word count is recalculated.  If there were not enough words then the UI won't change because it is already unanswered.

unanswered implies never answered or that the user is currently editing their answer.

freetextresponse/freetextresponse.py
caseylitton: removed, that was temporary

freetextresponse/freetextresponse.py
caseylitton: @stvstnfrd @caesar2164 

I'm leaning toward no change at this point.  It is not a necessary one.  I'll just add comments above the two lines in save and submit to make the relationships more clear.  

UPDATE:  While writing the comments, I realized that "has_remaining_attempts" changes value during a submit.  The suggested change is actually in reference to a FAIL SAFE.  If that line fails, then the UI was closed during the last submit.  Closing the UI happens in get_nodisplay_class(), which has the same line of conditionals.  This double check seems important so students can't over submit on graded problems, but the extra check causes confusion.  A more appropriate name for the proposed function is "was_allowed_to_submit()" or "can_save_reponse()".  "can_keep_attempting()" is actually is already written as "_get_nodisplay_class()". 

As a side note: save has the same was_allowed_to_submit because the save button is hidden along with the submission button.

freetextresponse/freetextresponse.py
caseylitton: Again, I didn't write this xblock.  I only assumed that it was done to limit the value.  I don't know why it was actually done this way.

freetextresponse/freetextresponse.py
stvstnfrd: > I'll just add comments above the two lines in save and submit to make the relationships more clear.

@caseylitton Any time you need to add comments to clarify a boolean expression, you have an over-complicated boolean expression that should be refactored.

As you point out, this conditional appears in multiple locations.
If we're duplicating the same expression 3 times,
we benefit from a helper method, especially here since it's purely state-based
(all checks can be determined from properties on the `self` object,
no outside arguments are passed); this feels to be a natural abstraction.

If we need to fix a bug or otherwise restructure the logic behind that check,
we're better served having it centralized in a single location.
Sure, some helpers are over-kill/premature, but this PR is already an example of us needing to change this exact check.

freetextresponse/private/view.js
stvstnfrd: You replied to a similar line previously; the comment appears to have been buried..

> I'm not following, this is consistent with the button click event callbacks.

In this one location, the form is `function()`;
in three other locations, the form is `function ()`.

I'm saying, at the very least, this should be consistent.
Given that named functions _must_ have a space after the `function` keyword anyway
(`function foo()`), we should do the same here so it's consistent everywhere: `function ()`.

Visual consistency aside, this is helpful for searching and other forms of pattern matching.

freetextresponse/freetextresponse.py
caseylitton: @stvstnfrd please see the updated comment.  Clarifying this boolean expression is needed.  We all keep forgetting that this conditional is a fail safe.  The value of it at the beginning of the function is different than at the end.

freetextresponse/freetextresponse.py
stvstnfrd: > Again, I didn't write this xblock.

I'm aware; I've acknowledged this elsewhere in the thread.

You've now refactored (nearly) this entire submit handler.
When the next person comes along to work on this repo, they're going to `git blame` these lines
and they'll resolve to you; it's your code now, not Azim's.

Look, I'm not blaming you; I'm pointing out that, in lieu of a counter-argument,
this code is wrong. We should fix it while we're editing it anyway.

```
# We don't need to conditionalize this at all based on `max_attempts`
if self._word_count_valid():
    self.count_attempts += 1
```

freetextresponse/private/view.js
caseylitton: Oh, Your talking about a space between the function and the parens.   It would have been more clear to explicitly write that out like, "you're missing a space between the function keyword and it's parentheses."  Fixed

freetextresponse/freetextresponse.py
stvstnfrd: @caseylitton Maybe we're talking about different things?

If `min_words` is 5, but my submitted response is only 3 words, my response is `incorrect`.
But the way I'm reading this logic, you only ever display "correct" or "incorrect" if
`min <= len(response) <= max`, right?

Or am I misreading this?

caseylitton: @stvstnfrd  You are reading the condition correctly, but responses that don't meet the minimum word count are invalid and therefore not accepted as a submission.  They are not scored(published) and count_attempts is not incremented.  This was original logic in our advanced xblock.   

I was curious so I just checked it out in studio.  Basic text responses allow users to submit empty responses for credit(which will always be incorrect) because instructors can't set the correct answer to an empty string.  I got an error in studio when I tried this.  

We could mimic this behavior, but then we'll need to display the min word count on the xblock prior to submission.  Currently we display the word requirements after submission.

freetextresponse/freetextresponse.py
caseylitton: I was just pointing out that I have no idea why the code was written this way because I didn't write it.   I'm not worried about integer overflow, I just thought that could have been a reason.    

I'm not altering production code with out a good reason.  If you can explain why pinning count_attempts is wrong then I will change it.

caseylitton:
@stvstnfrd @caesar2164 Some final changes to PR

stvstnfrd:
@caseylitton Please don't squash PRs mid-review;
it's impossible to determine what changed from one round to another.

This increases cognitive load as well as the amount of time needed for all reviewers.
It also leads to issues slipping through and/or comments not being addressed.

freetextresponse/freetextresponse.py
caesar2164: make sure you clean up the docstring!

freetextresponse/private/view.less
```diff
@@ -1,8 +1,8 @@
 .freetextresponse {
 
-  .user-input {
+  .user_input {
```
caesar2164, line 3: Wait, why are we changing the CSS to match the incorrectly named class? HTML classes should ALWAYS use dashes to split words.

> JS/jQuery: useCamelCase
> HTML: use-dashes
> Python: use_underscores

freetextresponse/private/view.less
```diff
@@ -24,13 +24,40 @@
       height: 20px;
       background: url("images/incorrect-icon.png") center center no-repeat;
     }
+
+    .student_answer {
```
caesar2164, line 28: same here, class should use dashes

freetextresponse/private/view.less
```diff
@@ -66,11 +93,14 @@
     display: inline-block;
   }
 
-  .word-count-error {
-    color: #A50000;
+  .capa_alert {
```
caesar2164, line 104: dashes

freetextresponse/private/view.html
```diff
@@ -1,24 +1,27 @@
-<div class="freetextresponse">
+<div class="freetextresponse xmodule_display xmodule_CapaModule problem">
```
caesar2164, line 1: use dashes for HTML classes

freetextresponse/private/view.html
```diff
@@ -1,24 +1,27 @@
...
     <h2 class="problem-header">{self.display_name}</h2>
     <div class="problem-progress">{problem_progress}</div>
     <p>{self.prompt}</p>
-    <div class="user-input {indicator_class}">
-        <textarea class="student_answer" style="height:150px" rows="20" cols="70">{self.student_answer}</textarea>
-        <span class="status {indicator_visibility_class}" aria-describedby="student_answer"></span>
+    <div class="capa_inputtype textline">
+        <div class="user_input {indicator_class}">
```
caesar2164, line 7: use dashes for HTML classes

freetextresponse/private/view.html
```diff
@@ -1,24 +1,27 @@
...
+            <textarea class="student_answer" rows="20" cols="70">{self.student_answer}</textarea>
+            <span class="status {visibility_class}" aria-describedby="student_answer"></span>
+        </div>
     </div>
-    <div class="submission-received">
+    <div class="capa_alert submission-received">
```
caesar2164, line 12: use dashes for HTML classes

freetextresponse/private/view.html
```diff
@@ -1,24 +1,27 @@
...
-    <div class="word-count-error">
-        {word_count_message}
-    </div>
     <div class="action">
-        <span class="{submit_class}">
-            <button class="check Submit" data-checking="Checking..." data-value="Submit" aria-disabled="false">
-                Submit
-            </button>
-        </span>
+        <button class="check Submit {nodisplay_class}" data-checking="Checking..." data-value="Submit">
```
caesar2164, line 16: this is the correct way to use a class variable. This way it's easy to tell that {nodisplay_class} comes from python...

freetextresponse/private/view.html
```diff
@@ -1,24 +1,27 @@
...
+            Submit
+        </button>
+        <button class="save {nodisplay_class}" data-value="Save">
+            Save
+        </button>
         <div class="used-attempts-feedback" aria-live="polite">{used_attempts_feedback}</div>
     </div>
+    <div class="capa_alert user_alert">
```
caesar2164, line 24: these should use dashes

freetextresponse/freetextresponse.py
stvstnfrd: > we'll need to display the min word count on the xblock prior to submission.

Yeah, that's what we need to do;
we can't just ignore invalid submissions.

freetextresponse/freetextresponse.py
stvstnfrd: Yeah, it looks like this still hasn't been removed.

freetextresponse/private/view.html
stvstnfrd, line 16: Just a sanity check: is the `Submit` class really supposed to be capitalized?

freetextresponse/private/view.js
```diff
@@ -46,4 +58,42 @@ function FreeTextResponseView(runtime, element) {
         });
         return false;
     });
+
+    buttonSave.on('click', function () {
+        buttonSave.text('Checking...');
```
stvstnfrd, line 63: This fine for now and I'm just thinking out loud here, but eventually
we need to start i18n/l10n'ing our XBlocks.

edX has done related work.

stvstnfrd (changes requested):
Mostly just questions/comments.

The one, major action item though is that we need to continue counting attempts as "invalid" if the word count is below the min threshold.

freetextresponse/private/view.html
caseylitton: This is in the openedx platform

freetextresponse/freetextresponse.py
caseylitton: We're now counting and grading all student answers that are submitted.

freetextresponse/freetextresponse.py
caseylitton: Resolved. moved back to top

freetextresponse/private/view.js
caseylitton: Resolved.

freetextresponse/freetextresponse.py
caseylitton: Resolved

caesar2164: This only gets executed when they click "Submit" right?
They can still "Save" even if their text is too short?

caseylitton: The indicator_class variable set by that function only impacts client side refresh and submit.   It is currently in the server side save function however and doesn't need to be.  I'll remove that.

freetextresponse/private/view.less
caseylitton: this is definitely edx.

freetextresponse/private/view.html
caseylitton: these are definitely edx.

freetextresponse/private/view.html
caseylitton: this is definitely edx.

freetextresponse/private/view.less
caseylitton: I'll check the dash but most are used by edx

tox.ini
```diff
@@ -40,5 +40,5 @@ setenv =
     NOSE_WITH_COVERAGE=1
 passenv = TRAVIS TRAVIS_JOB_ID TRAVIS_BRANCH
 commands =
-    {envbindir}/coverage run --source=freetextresponse manage.py test
+    {envbindir}/coverage run --source=freetextresponse manage.py test
```
caesar2164: don't add the space...

caesar2164:
as we discussed F2F, remove `freetextresponse/tests/test_travis.py`

freetextresponse/freetextresponse.py
caesar2164: cool, sounds good.

freetextresponse/private/view.html
caesar2164: sad, ok

freetextresponse/private/view.html
caesar2164: sad, ok

freetextresponse/private/view.html
caesar2164: probably from platform, disregard

freetextresponse/private/view.less
caesar2164: k, thats sad but fine

freetextresponse/private/view.less
caesar2164: k

caesar2164:
I'm good now, 👍

caesar2164 (approved):

freetextresponse/freetextresponse.py
```diff
@@ -35,8 +35,41 @@ def workbench_scenarios():
         scenarios = [
             ('Free-text Response XBlock',
              '''<sequence_demo>
-                    <freetextresponse />
-                    <freetextresponse name='My First XBlock' />
+                    <freetextresponse name='My First XBlock'/>
```
stvstnfrd: Let's just restore this space to remove the line from the diff.

freetextresponse/freetextresponse.py
stvstnfrd: I'm not seeing that line here. We should always provide an example of the default case, with no customization.

freetextresponse/tests/determine_credit.json
```diff
@@ -0,0 +1,75 @@
+{
```
stvstnfrd, line 1: Wait, where are these files being referenced?

stvstnfrd (changes requested):

freetextresponse/tests/determine_credit.json
caseylitton: All freetextreponse/tests/*.json are used in test.py for ddt test data.

stvstnfrd: Oh weird, Github was hiding the diff from me...
Thanks!

freetextresponse/tests.py
```diff
@@ -35,61 +61,165 @@ def make_an_xblock(cls, **kw):
...
+            word_count_message=self.xblock._get_word_count_message(),
+            indicator_class=self.xblock._get_indicator_class(),
+            problem_progress=self.xblock._get_problem_progress(),
+            used_attempts_feedback=self.xblock._get_used_attempts_feedback(),
+            nodisplay_class=self.xblock._get_nodisplay_class(),
+            visibility_class=self.xblock._get_indicator_visibility_class(),
+            submitted_message='',
+            user_alert='',
+        )
+        self.assertTrue(student_view_html == test_result)
```
stvstnfrd: Shouldn't this be `assertEquals`?

caseylitton: yes.  fixed now.

freetextresponse/tests.py
```diff
@@ -35,61 +61,165 @@ def make_an_xblock(cls, **kw):
...
+        test_data.weight = test_dict['weight']
+        test_data.max_attempts = test_dict['max_attempts']
+        test_data.max_word_count = test_dict['max_word_count']
+        test_data.min_word_count = test_dict['min_word_count']
+        test_data.submitted_message = test_dict['submitted_message']
+        validation = set()
+        self.xblock.validate_field_data(validation, test_data)
+        validation_list = list(validation)
+        # Only one validation error should be in set
+        self.assertTrue(len(validation_list) == 1)
```
stvstnfrd: Shouldn't this be `assertEquals`?

freetextresponse/tests.py
```diff
@@ -35,61 +61,165 @@ def make_an_xblock(cls, **kw):
...
+        public_path = '/resource/freetextresponse/public'
+        result = "{public_path}/{path}".format(
+            public_path=public_path,
+            path=path,
+        )
+        self.xblock.runtime.local_resource_url = MagicMock(
+            return_value=result
+        )
+        test_result = self.xblock.get_resource_url(path)
+        self.assertTrue(result == test_result)
```
stvstnfrd: Let's avoid this idiom anywhere it occurs.
If we're checking equality, always use `assertEquals`.

freetextresponse/tests.py
```diff
@@ -1,26 +1,52 @@
...
 from opaque_keys.edx.locations import SlashSeparatedCourseKey
 
 from xblock.field_data import DictFieldData
+from xblock.validation import ValidationMessage
 
 from .freetextresponse import Credit
 from .freetextresponse import FreeTextResponse
 
 
+class TestData(object):
```
stvstnfrd, line 21: Is there a reason this needed to be a new class instead of just a `dict`?

freetextresponse/tests.py
caseylitton: Oh, I wrote this so long ago.  Since then I've looked into the assert possibilities.  I'll recheck the tests.

stvstnfrd (approved):

PR #8, opened 2016-10-25
Remove Credit enumeration and clean up score calculation

kluo:
Submissions were failing to go through due to incompatible score type from the Credit enum. Removed and simplified some logic and tests.
-- UPDATE: kept Enum for code clarity, fixed to use Enum value.

@stvstnfrd

### DIFF
--- a/freetextresponse/freetextresponse.py
+++ b/freetextresponse/freetextresponse.py
@@ -394,13 +394,13 @@ class FreeTextResponse(StudioEditableXBlockMixin, XBlock):
         based on their answer
         """
         credit = self._determine_credit()
-        self.score = credit
+        self.score = credit.value
         self.runtime.publish(
             self,
             'grade',
             {
                 'value': self.score,
-                'max_value': Credit.full
+                'max_value': Credit.full.value
             }
         )
 
--- a/freetextresponse/tests.py
+++ b/freetextresponse/tests.py
@@ -392,7 +392,7 @@ class FreetextResponseXblockTestCase(unittest.TestCase):
         self.xblock.runtime.publish.assert_called_with(
             self.xblock,
             'grade',
-            {'value': Credit.full, 'max_value': Credit.full},
+            {'value': Credit.full.value, 'max_value': Credit.full.value},
         )
 
     def test_compute_score_half_credit(self):
@@ -411,7 +411,7 @@ class FreetextResponseXblockTestCase(unittest.TestCase):
         self.xblock.runtime.publish.assert_called_with(
             self.xblock,
             'grade',
-            {'value': Credit.half, 'max_value': Credit.full},
+            {'value': Credit.half.value, 'max_value': Credit.full.value},
         )
 
     def test_compute_score_no_credit(self):
@@ -430,7 +430,7 @@ class FreetextResponseXblockTestCase(unittest.TestCase):
         self.xblock.runtime.publish.assert_called_with(
             self.xblock,
             'grade',
-            {'value': Credit.zero, 'max_value': Credit.full},
+            {'value': Credit.zero.value, 'max_value': Credit.full.value},
         )
 
     def test_indicator_visibility_class_blank(self):


freetextresponse/freetextresponse.py
```diff
@@ -325,7 +324,7 @@ def _get_indicator_class(self):
         result = ''
         if self.count_attempts == 0:
             result = 'unanswered'
-        elif self._determine_credit() == Credit.zero:
+        elif self.score = 0:
```
stvstnfrd: Wait, that's gotta be `==`, right?

freetextresponse/freetextresponse.py
```diff
@@ -390,44 +389,26 @@ def _get_problem_progress(self):
...
-        """
-        result = None
-        if self.student_answer == '' or not self._word_count_valid():
-            result = Credit.zero
+        if self.student_answer == '':
+            self.score = 0
         elif not self.fullcredit_keyphrases \
                 and not self.halfcredit_keyphrases:
-            result = Credit.full
+            self.score = 1
```
stvstnfrd: Does this need to be `1.0`?

freetextresponse/freetextresponse.py
```diff
@@ -390,44 +389,26 @@ def _get_problem_progress(self):
...
         elif FreeTextResponse._is_at_least_one_phrase_present(
                 self.fullcredit_keyphrases,
                 self.student_answer
         ):
-            result = Credit.full
+            self.score = 1
```
stvstnfrd: Does this need to be `1.0`?

freetextresponse/freetextresponse.py
```diff
@@ -390,44 +389,26 @@ def _get_problem_progress(self):
...
         elif FreeTextResponse._is_at_least_one_phrase_present(
                 self.halfcredit_keyphrases,
                 self.student_answer
         ):
-            result = Credit.half
+            self.score = 0.5
         else:
-            result = Credit.zero
-        return result
+            self.score = 0
```
stvstnfrd: Does this need to be `0.0`?

freetextresponse/freetextresponse.py
```diff
@@ -484,6 +465,14 @@ def submit(self, data, suffix=''):
             else:
                 self.count_attempts += 1
             self._compute_score()
+            self.runtime.publish(
+                self,
+                'grade',
+                {
+                    'value': self.score,
+                    'max_value': 1
```
stvstnfrd: Does this need to be `1.0`?

freetextresponse/freetextresponse.py
```diff
@@ -484,6 +465,14 @@ def submit(self, data, suffix=''):
...
+                }
```
stvstnfrd: Trailing comma?

freetextresponse/tests.py
```diff
@@ -376,63 +375,6 @@ def test_problem_progress_score_positive_weight_plural(self):
             self.xblock._get_problem_progress(),
         )
 
-    def test_compute_score_full_credit(self):
```
stvstnfrd, line 379: Why are we dropping all of these?

freetextresponse/freetextresponse.py
```diff
@@ -496,14 +485,3 @@ def submit(self, data, suffix=''):
             'submitted_message': self._get_submitted_message(),
         }
         return result
-
-
-class Credit(Enum):
```
stvstnfrd, line 501: What went wrong here?

freetextresponse/freetextresponse.py
```diff
@@ -390,44 +389,26 @@ def _get_problem_progress(self):
 
     def _compute_score(self):
         """
-        Computes and publishes the user's core for the XBlock
-        based on their answer
+        Assuming student answer has valid word count, computes the score for
```
stvstnfrd: Where is that assumption verified/enforced?

stvstnfrd (changes requested):
Can you provide a little more background on why we needed this invasive of an approach?
What went wrong with the Enum?

freetextresponse/freetextresponse.py
kluo: My change to the score to directly use the Credit values didn't play nice when publishing the grade, it causes django to throw
`TypeError: float() argument must be a string or a number`
when trying to save the score for the student_module. From looking into the code here, there's no real reason we ever needed this extra Enum, just makes things more clunky from my view. So I removed it and the tests related to it, and just updated compute_score

freetextresponse/tests.py
kluo: See above

freetextresponse/freetextresponse.py
kluo: In the submit handler

freetextresponse/freetextresponse.py
kluo: Not needed as long as the value can be parsed as float, it won't cause errors

freetextresponse/freetextresponse.py
kluo: See above

freetextresponse/freetextresponse.py
kluo: See above

freetextresponse/freetextresponse.py
kluo: See above

freetextresponse/freetextresponse.py
stvstnfrd: While not strictly required here, I'm generally a fan of enums and constants for values like this. The main benefits are:
- easier to read
  - it's immediately obvious what the effect of `Credit.full` vs. `Credit.half` is, without needing to figure out what the scale is
- easier to update
  - just change the values once, instead of everywhere they occur
    - right now the code is able to cast those integers `1` and `0` to `float`, but what if the grading functionality is changed to require `float`s to be passed in? Then we're back to tracking down every single occurrence; this is error prone.
- easier to trace/search
  - if I want to track down all credit-related code paths, I could (prior to this) search for `Credit`; if we change it to literal values, you risk finding many false positives, as there could be plenty of lines that contain the number `0`.

I think you may have overthought this a bit;
I can see two simpler, less invasive approaches:
- just cast from enum to float where needed (ie, the call to grade.publish)
- turn those enums into `float` constants
  - this gives us the same benefits; single definition, human-readable name, etc, but allows us to maintain the intended data type (`float`)

``` python
CREDIT_FULL = 1.0
CREDIT_HALF = 0.5
CREDIT_NONE = 0.0
```

stvstnfrd (approved):
+1

PR #10, opened 2016-12-01
Testing tox

caseylitton:
Updates for running tox tests on a local devstack.  Checking if they work in Travis


caseylitton:
@stvstnfrd @kluo @caesar2164 PR to fix travis tests on Kevin's last commit.  The updates were based 
https://github.com/Stanford-Online/xblock-in-video-quiz 
but needed some tweaks.  This setup also passes local tox testing on a devstack.

requirements.txt
```diff
@@ -0,0 +1,4 @@
+git+https://github.com/edx/XBlock.git#egg=XBlock
```
caesar2164, line 2: enum34
django

caesar2164: put those above this line...

tox.ini
```diff
@@ -8,8 +8,6 @@ commands = {envpython} manage.py test
 [testenv:py27-dj18]
 deps =
     -rrequirements.txt
-    enum34
-    django >= 1.8, < 1.9
```
caesar2164: this line should stay

requirements.txt
```diff
@@ -1,3 +1,5 @@
+django<1.9,>=1.8
```
caesar2164, line 1: this should just be "django"

stvstnfrd: Should it? That means it will install the newest version.

caesar2164: yeah, cause we don't care which version of django it uses except for the specific DJ1.8 test env

tox.ini
```diff
@@ -0,0 +1,44 @@
+[tox]
+downloadcache = {toxworkdir}/_download/
+envlist = py27-dj18,coverage,pep8,pylint
+
+[testenv]
+commands = {envpython} manage.py test
+
+[testenv:py27-dj18]
+deps =
+    -rrequirements.txt
```
caesar2164, line 10: there should be a specific `django<1.9,>=1.8` here

requirements.txt
caseylitton: 1. In the tox.ini in three places 
Or 
2. Once in requirements and then gets installed for pep8 and pylint when it's not needed.

(1) is more efficient for loading and (2) is more efficient for running.  I'd opt to put it in the needed test in the tox.ini

requirements.txt
caseylitton: django was moved in from tox.ini and enum34 was add to setup.py.  Adding django to setup.py did not help.

tox.ini
caseylitton: 1. In the tox.ini in three places 
Or 
2. Once in requirements and then gets installed for pep8 and pylint when it's not needed.

(1) is more efficient for loading and (2) is more efficient for running.  I'd opt to put it in the needed test in the tox.ini

freetextresponse/freetextresponse.py
```diff
@@ -405,6 +407,7 @@ def _compute_score(self):
         )
 
     def _determine_credit(self):
+        # pylint: disable=redefined-variable-type
```
caesar2164, line 412: this seems like a failure we could fix, instead of just disabling it

caesar2164:
this looks really good to me, I'd love some background on the disabled pylint tests tho...

freetextresponse/freetextresponse.py
caseylitton: This is an issue with how pylint handles enums/conditionals.   I think it is fixed in a newer version.

Redefinition of result type from freetextresponse.freetextresponse.Credit.zero to freetextresponse.freetextresponse.Credit.full

caseylitton: This is still an issue in pylint 1.6.4

freetextresponse/tests.py
```diff
@@ -3,11 +3,14 @@
 """
 import unittest
 
-import mock
 from django.test.client import Client
 from django.utils.translation import ugettext as _
+
+import mock
 from mock import MagicMock
```
dcadams: Do you need both of these imports?

setup.py
```diff
@@ -1,24 +1,50 @@
 import json
 import setuptools
+from setuptools.command.test import test as TestCommand
```
dcadams, line 3: Do you need both of these imports?

dcadams:
:+1:

caesar2164:
👍

PR #11, opened 2017-01-26
Add Mobile Support

libremente:
This PR introduces the following:

- Mobile Compatibility by means of adding the proper decorator to the student_view
- Fixed the appearance of the input textarea by removing the cols property previously embedded in the view.html and now styled by means of `width` property inside the view.less file
- Updates the files in the `public` dir generated by grunt

Now in the app the view is the following:
screenshot_2017-01-25-15-17-10_org edx mobile

@caesar2164 @stvstnfrd @caseylitton


stvstnfrd:
@libremente This looks good to me. Out of curiosity, what did the textarea look like before?

libremente:
> This looks good to me. Out of curiosity, what did the textarea look like before?

@stvstnfrd It was fixed spanning over 70 cols so in the mobile app (or also re-dimensioning the browser window) it was going out of scope and it required to use the horizontal scroll. I believe now it's more "usable".

caseylitton:
+1

stvstnfrd:
+1

PR #12, opened 2017-02-08
Fix progress page max_score scoring issue

caseylitton:
max_score function needs to be included in
xblocks so the LMS progress page correctly
reports the progress for a subsection. Not
fixing this is confusing to the student and
could result in grading errors.

@kluo @stvstnfrd PR for fixing progress issue by adding max_score function to xblock.

https://trello.com/c/PqrXDAYi/669-graded-xblocks-not-correctly-reflected-in-grading

### DIFF
--- a/freetextresponse/freetextresponse.py
+++ b/freetextresponse/freetextresponse.py
@@ -246,6 +246,16 @@ class FreeTextResponse(StudioEditableXBlockMixin, XBlock):
         )
         return fragment
 
+    def max_score(self):
+        """
+        Returns the configured number of possible points for this component.
+        Arguments:
+            None
+        Returns:
+            float: The number of possible points for this component
+        """
+        return self.weight
+
     @classmethod
     def _generate_validation_message(cls, msg):
         """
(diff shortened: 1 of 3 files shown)


freetextresponse/freetextresponse.py
```diff
@@ -242,6 +242,16 @@ def student_view(self, context=None):
...
 
+    def max_score(self):
+        """
+        Returns the configured number of possible points for this component.
+        Arguments:
+            None
+        Returns:
+            float: The number of possible points for this component
+        """
+        return self.weight if self.has_score else None
```
kluo: Shouldn't this always return the _raw_ max score, `Credit.full.value`? The `has_score` attribute is defaulted to True so the boolean check is superfluous.

kluo (changes requested):

freetextresponse/freetextresponse.py
caseylitton: Based on other CAPA xblocks and their progress page, this should return the instructor weight.  Credit.full.value is always one.  If the weight is 3 for an xblock then the progress page available points goes up by 3 for that subsection.  

Per discussion, I'll update so max_score returns self.weight with no checks.

stvstnfrd:
This seems reasonable, +1

kluo:
+1

PR #14, opened 2017-04-13
Add hints feedback

caseylitton:
@stvstnfrd PR for Hint and Feedback features for free text response.  

Hints and feedback are separate features.  The feedback requires the more complex dictionary list we've seen in the adaptive feedback.  These dicts are simpler but different.  'phrase' and 'feedback' and the dict keys.  So if the 'phrase' is found in the free text response the user will receive the associate feedback.  Original scoring ideals were not changed, except zero credit dicts were added to allow for feedback on common wrong answers.


stvstnfrd:
@caseylitton Github is complaining about conflicts?

stvstnfrd:
@caseylitton Tests are failing on travis-ci.
Can you take a look?

PR #15, opened 2017-06-12
Fix untranslated strings for know hbv vietnamese

caseylitton:
@stvstnfrd @kluo PR to fix certain translations for the free-text-response xblock.   Studio settings translation hooks in the python code had to be removed.  See image below for error.   Also this xblock would need to be added to cms and/or lms OPTIONAL_APPS in platform just like ora2.

In order to finish this PR I need the translation files with the following two translations

"Your response must be between 1 and 10000 words" - where 1 and 1000 are variables and 'words' is singular and plural.

"Invalid Word Count. Your response must be between 1 and 10000 words."  - same as first one

This PR does not fix three issues associated with Button texts. Submit, Save and 'Checking...'.  In order to translate those we would need to change from mako to django templates.


caseylitton:
Error that prompted removing python translations to studio settings.

screen shot 2017-06-08 at 12 28 51 pm

stvstnfrd:
@caseylitton Re: the studio error; I think we just want to use the lazy translation for any strings that are executed at import time. See [1].

[1] https://docs.djangoproject.com/en/1.11/topics/i18n/translation/#lazy-translations

conf/locale
```diff
@@ -0,0 +1 @@
+../freetextresponse/locale/
```
stvstnfrd, line 1: Why the symlink? Why do we need both paths?

freetextresponse/locale/vi/LC_MESSAGES/django.po
```diff
@@ -0,0 +1,41 @@
+# PLACE HOLDER FILE, 
+# only the mo is used but this is where I found a translated string to test.
```
stvstnfrd: What do you mean by "placeholder file"?
`.mo` files are the binary translation file that is used at runtime,
but it's created _from_ the `.po` file.

freetextresponse/freetextresponse.py
```diff
@@ -178,10 +178,8 @@ def workbench_scenarios():
             'This is the message students will see upon '
             'submitting a draft response'
         ),
-        default=_(
-            'Your answers have been saved but not graded. '
-            'Click "Submit" to grade them.'
-        ),
+        default='Your answers have been saved but not graded. '
```
stvstnfrd: Let's at least still wrap this in parens;
the indenting is wonky with the implicit line continuation.

conf/locale
caseylitton: Eventually we found a reason for this.  Translations were not found without adding xblock plath to LOCALE_PATHS if the symlink was not in place.  Symlink allows us to only add the xblock to OPTIONAL_APPS

caseylitton:
The failing test is due to the test infrastructure not being able to find the html template when django.template.loader import get_template function is called.  We previously did not use this function because we loaded the file directly.  I'm working on a solution for that now.  I assume we can move on without the fix however?

freetextresponse/freetextresponse.py
```diff
@@ -211,38 +210,76 @@ def workbench_scenarios():
...
+        context_dict = {
+            'display_name': self.display_name,
+            'indicator_class': self._get_indicator_class(),
+            'nodisplay_class': self._get_nodisplay_class(),
+            'problem_progress': self._get_problem_progress(),
+            'prompt': self.prompt,
+            'student_answer': self.student_answer,
+            # 'submitted_message': self._get_submitted_message(),
+            'used_attempts_feedback': self._get_used_attempts_feedback(),
+            # 'user_alert': '',
```
stvstnfrd: Should this be deleted?

freetextresponse/freetextresponse.py
```diff
@@ -211,38 +210,76 @@ def workbench_scenarios():
...
+        # All data we intend to pass to the front end.
+        self.runtime.service(self, 'i18n')
+        context_dict = {
+            'display_name': self.display_name,
+            'indicator_class': self._get_indicator_class(),
+            'nodisplay_class': self._get_nodisplay_class(),
+            'problem_progress': self._get_problem_progress(),
+            'prompt': self.prompt,
+            'student_answer': self.student_answer,
+            # 'submitted_message': self._get_submitted_message(),
```
stvstnfrd: Should this be deleted?

freetextresponse/freetextresponse.py
```diff
@@ -211,38 +210,76 @@ def workbench_scenarios():
...
+        template = get_template("freetextresponse_view.html")
         fragment = self.build_fragment(
-            html_source=view_html,
-            paths_css=[
-                'view.less.min.css',
+            template,
+            context_dict,
+            initialize_js_func='FreeTextResponseView',
+            additional_css=[
+                'public/view.less.min.css'
```
stvstnfrd: Add trailing comma.

freetextresponse/freetextresponse.py
```diff
@@ -211,38 +210,76 @@ def workbench_scenarios():
...
             ],
-            paths_js=[
-                'view.js.min.js',
+            additional_js=[
+                'public/view.js.min.js'
```
stvstnfrd: Add trailing comma.

package.json
```diff
@@ -0,0 +1,38 @@
+{
```
stvstnfrd, line 1: Had we gotten rid of this file?

freetextresponse/tests.py
```diff
@@ -118,103 +118,6 @@ def test_validate_field_data(self, **test_dict):
             validation_list[0].text,
         )
 
-    def test_get_resource_string(self):
```
stvstnfrd, line 121: Why are we deleting this block of tests without replacing the functionality?

freetextresponse/templates/freetextresponse_view.html
```diff
@@ -0,0 +1,29 @@
...
+        <button class="save {{ nodisplay_class }}" data-checking="{% trans "Checking..."%}" data-value={% trans "Save" %}>
+            {% trans "Save" %}
+        </button>
+        <div class="used-attempts-feedback" aria-live="polite">{{ used_attempts_feedback }}</div>
+    </div>
+    {% if user_alert %}
+        <div class="capa_alert user_alert">{{ user_alert }}</div>
+    {% endif %}
+</div>
+
```
stvstnfrd: This file ends in one too many newlines.

freetextresponse/templates/freetextresponse_view.html
```diff
@@ -0,0 +1,29 @@
...
+        </div>
+    </div>
+    {% if submitted_message %}
+        <div class="capa_alert submission-received">{{ submitted_message }}</div>
+    {% endif %}
+    <div class="action">
+        <button class="check Submit {{ nodisplay_class }}" data-checking="{% trans "Checking..."%}" data-value={% trans "Submit" %}>
+            {% trans "Submit" %}
+        </button>
+        <button class="save {{ nodisplay_class }}" data-checking="{% trans "Checking..."%}" data-value={% trans "Save" %}>
```
stvstnfrd: Add a space between `"Checking..."` and `%}`,
per the convention you're using elsewhere.

freetextresponse/templates/freetextresponse_view.html
```diff
@@ -0,0 +1,29 @@
...
+        <div class="user_input {{ indicator_class }}">
+            <textarea class="student_answer" rows="20">{{ student_answer }}</textarea>
+            <span class="status {{ visibility_class }}" aria-describedby="student_answer"></span>
+        </div>
+    </div>
+    {% if submitted_message %}
+        <div class="capa_alert submission-received">{{ submitted_message }}</div>
+    {% endif %}
+    <div class="action">
+        <button class="check Submit {{ nodisplay_class }}" data-checking="{% trans "Checking..."%}" data-value={% trans "Submit" %}>
```
stvstnfrd: Add a space between `"Checking..."` and `%}`,
per the convention you're using elsewhere.

freetextresponse/locale/vi/LC_MESSAGES/django.po
```diff
@@ -0,0 +1,48 @@
...
+# This file is distributed under the same license as the PACKAGE package.
+# FIRST AUTHOR <EMAIL@ADDRESS>, YEAR.
+# 
+#, fuzzy
+msgid ""
+msgstr ""
+"Project-Id-Version: PACKAGE VERSION\n"
+"Report-Msgid-Bugs-To: \n"
+"POT-Creation-Date: 2017-06-13 22:13+0000\n"
+"PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
```
stvstnfrd: This looks like it should be updated?

freetextresponse/locale/vi/LC_MESSAGES/django.po
```diff
@@ -0,0 +1,48 @@
+# Vietnamese Translations for Free Text Response Xblock
+# plural translations are copies of the singular version
+# Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
+# This file is distributed under the same license as the PACKAGE package.
+# FIRST AUTHOR <EMAIL@ADDRESS>, YEAR.
```
stvstnfrd: This whole header needs updated.

freetextresponse/freetextresponse.py
```diff
@@ -211,38 +210,76 @@ def workbench_scenarios():
...
-            used_attempts_feedback=self._get_used_attempts_feedback(),
-            nodisplay_class=self._get_nodisplay_class(),
-            visibility_class=self._get_indicator_visibility_class(),
-            submitted_message='',
-            user_alert='',
-        )
+
+        # All data we intend to pass to the front end.
+        self.runtime.service(self, 'i18n')
+        context_dict = {
```
stvstnfrd: By convention, this should just be called `context`;
we don't need to embed the datatype.

freetextresponse/freetextresponse.py
```diff
@@ -211,38 +210,76 @@ def workbench_scenarios():
...
-            indicator_class=self._get_indicator_class(),
-            problem_progress=self._get_problem_progress(),
-            used_attempts_feedback=self._get_used_attempts_feedback(),
-            nodisplay_class=self._get_nodisplay_class(),
-            visibility_class=self._get_indicator_visibility_class(),
-            submitted_message='',
-            user_alert='',
-        )
+
+        # All data we intend to pass to the front end.
```
stvstnfrd: We don't need to clarify what `context` means here,
as it's a standard `django-ism`.
We can remove this comment.

freetextresponse/freetextresponse.py
```diff
@@ -211,38 +210,76 @@ def workbench_scenarios():
...
+            'problem_progress': self._get_problem_progress(),
+            'prompt': self.prompt,
+            'student_answer': self.student_answer,
+            # 'submitted_message': self._get_submitted_message(),
+            'used_attempts_feedback': self._get_used_attempts_feedback(),
+            # 'user_alert': '',
+            'visibility_class': self._get_indicator_visibility_class(),
+            'word_count_message': self._get_word_count_message(),
+        }
+        template = get_template("freetextresponse_view.html")
```
stvstnfrd: Use single quotes, as elsewhere.

freetextresponse/freetextresponse.py
stvstnfrd: Take a look at the context being passed in to this method;
we may want to add to that context, instead of creating a new dict.

freetextresponse/freetextresponse.py
```diff
@@ -211,38 +210,76 @@ def workbench_scenarios():
...
     # section 'View @supports(multi_device) decorator'
     @XBlock.supports('multi_device')
     def student_view(self, context=None):
         # pylint: disable=unused-argument
+        """The main view of FreeTextResponse, displayed when viewing courses.
+
+        The main view which displays the general layout for FreeTextResponse
+
+        Args:
+            context: Not used for this view.
```
stvstnfrd, line 250: Now that we're templating, we probably should be using this.

stvstnfrd (changes requested):

package.json
caseylitton: it's needed for grunt, I've had to add this back a few times because it never got checked back in.

freetextresponse/tests.py
caseylitton: This series of functions was removed because the functionality was folded into build fragment.  Build fragment was removed b/c of an issue with the change to django template as seen in the failing test.    I'll ping you in Slack.

freetextresponse/freetextresponse.py
```diff
@@ -159,7 +158,7 @@ def workbench_scenarios():
             'This is the message students will see upon '
             'submitting their response'
         ),
-        default=_('Your submission has been received'),
```
caesar2164, line 162: why are we untranslating stuff?

freetextresponse/public/view.html
```diff
@@ -1 +0,0 @@
-<div class="freetextresponse xmodule_display xmodule_CapaModule problem"><h2 class="problem-header">{self.display_name}</h2><div class="problem-progress">{problem_progress}</div><p>{self.prompt}</p><div class="word-count-message">{word_count_message}</div><div class="capa_inputtype textline"><div class="user_input {indicator_class}"><textarea class="student_answer" rows="20">{self.student_answer}</textarea><span class="status {visibility_class}" aria-describedby="student_answer"></span></div></div><div class="capa_alert submission-received">{submitted_message}</div><div class="action"><button class="check Submit {nodisplay_class}" data-checking="Checking..." data-value="Submit">Submit</button> <button class="save {nodisplay_class}" data-value="Save">Save</button><div class="used-attempts-feedback" aria-live="polite">{used_attempts_feedback}</div></div><div class="capa_alert user_alert">{user_alert}</div></div>
```
caesar2164, line 1: I don't think you shouldn't be removing this file...

freetextresponse/freetextresponse.py
caseylitton: That was a weird issue, if we try to translate default='str' in the xblock settings variable it blows up on compiling.  See above error png.

freetextresponse/public/view.html
caseylitton: the file moved.  It must be in templates directory.

caseylitton: * in order to be found by get_template.  We moved to django templates.

setup.py
```diff
@@ -38,7 +38,6 @@ def run_tests(self):
         'django_nose',
         'mock',
         'coverage',
-        'mako',
```
kluo: Seems like we need to keep this because xblock-utils needs it (per broken travis build logs)

kluo (changes requested):

freetextresponse/freetextresponse.py
caesar2164: ah ok

freetextresponse/public/view.html
caesar2164: ok, thanks for explanation

caesar2164 (approved):
I'm 👍

PR #17, opened 2017-06-22
Fix prompt html support

caseylitton:
@kluo @caesar2164 PR to fix TWO bugs created when we switched to django templates for this xblock.

The first bug is a development one.  grunt no longer needs to watch the html file but I left the dir in the Gruntfile.

Second is a UI bug fix from the trello card, 

https://trello.com/c/41a2vUOR/903-fix-prompt-html-support-for-free-text-response-the-free-text-response-xblock-prompt-field-used-to-support-html-but-now-is-displa

Instructors previously added html to the prompt in Studio settings.  Django template variables need a autoescape tag 'safe' added to all html. 
 Django docs: https://docs.djangoproject.com/en/dev/ref/templates/builtins/

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -120,7 +120,6 @@ module.exports = function (grunt) {
                     gruntFile,
                     directoryPrivateJsAll,
                     directoryPrivateLessAll,
-                    directoryPrivateHtmlAll,
                 ],
                 tasks: [
                     'default',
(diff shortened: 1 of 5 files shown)


caesar2164:
hmm, can we add tests for the UI bug so we don't break it in future?

caesar2164:
killer! 👍

dcadams:
:+1:

PR #18, opened 2017-07-19
Add cache to display answer after submit or save

potsui:
@caseylitton @caesar2164 @stvstnfrd PR for caching answer when returning to a unit in current subsection with a free text response xblock

Previously in the free-text response xblock, when a user submits or
saves an answer, navigates to another page, and navigates back to the
original page, the user's answer disappears. Refreshing the webpage will
restore the answer, but this was nonetheless a disconcerting experience
for the user.

This fix caches the user's answer to display after submit or save, so
that when the user navigates away and back, the user can still view
their answer without having to manually refresh the page.

References:

- Fix carried over from the [submit and compare xblock](https://github.com/Stanford-Online/xblock-submit-and-compare/blob/master/submit_and_compare/static/js/submit_and_compare_view.js#L29).
- [Trello](https://trello.com/c/a6mYFMUu/935-from-participants-in-htlm-improve-experience-encountered-when-returning-to-a-unit-in-your-current-subsection-with-a-free-text-re)

Some issues that arose:

- Couldn't find any existing JS tests for the xblock. Just ran the existing python tests and checked in browser, but didn't write any unit tests checking this fix.
- I ran `npm install` and was still missing modules `has-color`, `less`, and `zlib-browserify`, so I manually `npm install --save`d them. It seems`package-lock.json` should be updated?
- Running `grunt watch` aborted on me due to errors: 
screen shot 2017-07-19 at 11 56 02 am I worked around this by specifying only the `concat` and `uglify` tasks to run (instead of `default` all) on `watch`, since I only changed a js file.

I may have messed around with my local environment a bit, so maybe these issues are only happening locally for my free text response xblock instance.

CC: @jspayd @jlikhuva


caseylitton:
Fix for the jshint issue in comments of Josh's PR.  https://github.com/Stanford-Online/xblock-qualtrics-survey/pull/6

We'll do two commits on this PR, the first one should be fixing these grunt issues.

potsui:
@caseylitton @caesar2164 @stvstnfrd resolved all grunt errors and included public files. The `jshint` error was resolved through Casey's comment above, `csslint` warnings can be ignored, and the rest of the issues were simply because I was using the latest version of `npm`.

For reference, this is currently the expected output of running `grunt watch`:
screen shot 2017-07-20 at 2 41 39 pm

caseylitton:
I checked your code on my devstack.  It seems to work but I found a preexisting bug.

Stop correctness indicator from showing on save
Reproduce:  
1. Create a basic free text response problem with default settings in studio
2. View it live in LMS
3. Save an answer.
4. Refresh browser.
5. See that the saved answer has the green correct indicator.

In freetextresponse.py there is a _get_indicator_class function that decides which css class the answer gets; unanswered, correct, or incorrect.  It sets the UI to display red, green or no colors.  There are a few conditions in that function.  I think the one causing the issue is the _determine_credit function.  That function can return Credit.zero,  Credit.half, Credit.full and None.  _get_indicator_class is not handling the None state correctly.  If credit is None, that implies that no answers have been submitted and an unanswered correctness indicator should be returned.  

Can you verify this bug?  And then check what happens if a you submit an answer and then save an answer.  I suspect we'll see the same problem because credit is permanently set to a non-None value after the first submission.  

If so, The only solution I can think of off hand is to add a flag that keeps track of whether the student is in the submit or save state.  Which could toggle the indicator to unanswered on that condition.  Unless you can think of something simpler we'll probably just make a new trello card for this bug.

Note for later(to add to trello card) that Monica brought up when I showed here this bug:  What do we do if a student saves and answer but never submits?  Should we automatically use their last save as a submission when the due date arrives?

Per Monica's request, I verifed the bug on production.

potsui:
The same thing happens on mine too. Furthermore, if you submit and then save another answer, and navigate forward and back (without refresh), the correctness indictor reverts to the last submitted answer. Refreshing yields the same bug as @caseylitton described. The submit/save flag sounds fine to me. 

One thought for the last issue: if a student saves and never submits any answer, it wouldn't hurt to use the saved answer as the submission. However, if a student has already submitted an answer and saves another, there are two cases that lead to divergent paths:

1. The student wants to use that saved answer to replace the previously submitted one, but simply forgot to submit. In this case we would want to auto submit.
2. The saved answer is just an accident or abandoned second attempt. In this case we would not want to auto submit.

An option for auto submit would be using the higher score of the saved and submitted answers, but this could potentially be exploited by students trying to maximize their score. Or maybe it would be better to somehow leave the decision making to the user (i.e. send a reminder for them to manually submit any saved answers)? @mondiaz

caseylitton:
Thanks for checking that other bug.  I made a trello card in the backlog for next sprint with our comments.  

The PR looks good, I tested it locally.   +1

Remember to squash the public js files into your feature commit after you get another plus one. 

bug @caesar2164

stvstnfrd:
Yeah, I was going to write about the issue of saving following submission; thanks @potsui .

I took a look at how this is handled by the existing text input problem.
It actually removes the `Save` button once you've hit `Submit`.
A user may reset/resubmit, but cannot save again after submission (unless they reset the problem).

This seems like the best approach,
as it's not surprising and is consistent with other text problems on the platform.

freetextresponse/private/view.js
```diff
@@ -15,6 +15,16 @@ function FreeTextResponseView(runtime, element) {
     var url = runtime.handlerUrl(element, 'submit');
     var urlSave = runtime.handlerUrl(element, 'save_reponse');
 
+    var xblockId = $element.attr('data-usage-id');
+    var cachedAnswerId = xblockId + '_cached_answer';
+    var problemProgressId = xblockId + '_problem_progress';
+    var usedAttemptsFeedbackId = xblockId + '_used_attempts_feedback';
+    if ($('body').data(cachedAnswerId) !== undefined) {
```
caesar2164, line 22: I don't love having all of the values hooked onto the `<body>` element on the page. Since we have access to `$element` let's bind the values to that, so each xblock has it's own data attached...

ie:

```javascript
if ($element.data(cachedAnswerId) !== undefined) {
    // Your code here...
}
```

freetextresponse/private/view.js
```diff
@@ -47,6 +57,10 @@ function FreeTextResponseView(runtime, element) {
                 userAlertMessage.text(response.user_alert);
                 buttonSave.addClass(response.nodisplay_class);
                 setClassForTextAreaParent(response.indicator_class); 
+
+                $('body').data(cachedAnswerId, $element.find('.student_answer').val());
+                $('body').data(problemProgressId, response.problem_progress);
+                $('body').data(usedAttemptsFeedbackId, response.used_attempts_feedback);
```
caesar2164, line 63: from my above comment, these would become:
```javascript
$element.data(cachedAnswerId, $element.find('.student_answer').val());
$element.data(problemProgressId, response.problem_progress);
$element.data(usedAttemptsFeedbackId, response.used_attempts_feedback);
```

freetextresponse/private/view.js
```diff
@@ -79,6 +93,10 @@ function FreeTextResponseView(runtime, element) {
                 buttonSave.text(buttonSave[0].dataset.value);
                 userAlertMessage.text(response.user_alert);
 
+                $('body').data(cachedAnswerId, $element.find('.student_answer').val());
+                $('body').data(problemProgressId, response.problem_progress);
+                $('body').data(usedAttemptsFeedbackId, response.used_attempts_feedback);
```
caesar2164, line 98: change these as well.

freetextresponse/public/view.js
```diff
@@ -15,6 +15,16 @@ function FreeTextResponseView(runtime, element) {
     var url = runtime.handlerUrl(element, 'submit');
     var urlSave = runtime.handlerUrl(element, 'save_reponse');
 
+    var xblockId = $element.attr('data-usage-id');
+    var cachedAnswerId = xblockId + '_cached_answer';
+    var problemProgressId = xblockId + '_problem_progress';
+    var usedAttemptsFeedbackId = xblockId + '_used_attempts_feedback';
+    if ($('body').data(cachedAnswerId) !== undefined) {
```
caesar2164, line 22: this file will get changed automagically once you edit the private version!

caesar2164 (changes requested):

potsui:
@caesar2164 I've made the changes as requested but it doesn't seem to be working, am currently investigating

potsui:
Per f2f w/ @caesar2164 , decided to undo suggested changes

freetextresponse/private/view.js
caesar2164: this doesn't work because `$element` is redefined on every run...

freetextresponse/private/view.js
caesar2164: this doesn't work because `$element` is redefined on every run...

freetextresponse/private/view.js
caesar2164: this doesn't work because `$element` is redefined on every run...

freetextresponse/public/view.js
caesar2164: this doesn't work because `$element` is redefined on every run...

caesar2164 (approved):

potsui:
@caesar2164 @caseylitton I've squashed the public files into feature commit

stvstnfrd:
@caesar2164 @potsui:
I'm not keen on caching data in the `<BODY>` tag either;
it's always possible that a future update or rogue script could
wipe/hijack data attributes from the `<BODY>` tag.
Plus it's just "polite" for our XBlock to keep to itself :)

The thing I see missing from the original implementation [1] is that the
element itself is never initialized on-load. If we added this directly
to the template [2] wouldn't that ensure it's always set for us, even if
the `$element is redefined on every run`? I'm imagining something like this:

```html
<div class="freetextresponse ..." data-cached-answer="{{ student_answer }}">
```

Or maybe I'm missing something from the F2F.
Either way, let me know what you think!

[1] https://github.com/Stanford-Online/xblock-free-text-response/commit/0d1a81a
[2] https://github.com/Stanford-Online/xblock-free-text-response/blob/master/freetextresponse/templates/freetextresponse_view.html#L2

potsui:
@stvstnfrd I'm looking more into it right now, will let you know!

mondiaz:
@potsui Sorry for the very belated response on my part. To your question from the other week:

My input would be that we should probably be consistent with whatever happens in capa problems, which I think is: someone saves a capa problem but does not submit it, and so does not get credit for it. So my vote would be for us to not submit a saved answer, only because it might create a confusing experience.

Thanks!

PR #19, opened 2017-07-21
Suppress IntegrityError exception

jlikhuva:
Added code to catch – and ignore – The
IntegrityError exception. Created a new
unit test that mocks the exception
in question to ensure that it is
gracefully ignored.

@caseylitton  @potsui @jspayd

### Release Notes
---

* Previously, a call to _compute_score could potentially yield an IntegrityError exception that had no handler.
* This change fixes that by wrapping a segment of _compute_score in a try-except and ignoring the IntegrityError exception.
* The change also includes unit tests to ensure that the exception is indeed being handled.

### DIFF
--- a/freetextresponse/tests.py
+++ b/freetextresponse/tests.py
@@ -12,6 +12,7 @@ from opaque_keys.edx.locations import SlashSeparatedCourseKey
 from xblock.field_data import DictFieldData
 from xblock.validation import ValidationMessage
 
+from django.db import IntegrityError
 from django.template.loader import get_template
 
 from .freetextresponse import Credit
@@ -259,6 +260,24 @@ class FreetextResponseXblockTestCase(unittest.TestCase):
             'grade',
             {'value': credit.value, 'max_value': Credit.full.value},
         )
+
+    def test_compute_score_integrity_error(self):
+        # pylint: disable=protected-access, invalid-name
+        """
+        Tests that _compute_score gracefully handles IntegrityError exception.
+
+        Tests to ensure that if an IntegrityError exception
+        is thrown by any of the methods/functions called in the
+        process of saving the score, the program handles it gracefully.
+        We force runtime.publish to throw an IntegrityError exception,
+        and expect _compute_score to be graceful about it.
+        """
+        self.xblock.runtime.publish = MagicMock(return_value=None)
+        self.xblock.runtime.publish.side_effect = IntegrityError(
+            "Unique Key Violation"
+        )
+        self.xblock._determine_credit = MagicMock(return_value=Credit.zero)
+        self.xblock._compute_score()
 
     def test_is_at_least_one_phrase_present(self):
         # pylint: disable=invalid-name, protected-access
--- a/freetextresponse/freetextresponse.py
+++ b/freetextresponse/freetextresponse.py
@@ -2,6 +2,7 @@
 This is the core logic for the Free-text Response XBlock
 """
 from enum import Enum
+from django.db import IntegrityError
 from django.template.context import Context
 from django.template.loader import get_template
 from django.utils.translation import ungettext
@@ -454,14 +455,17 @@ class FreeTextResponse(StudioEditableXBlockMixin, XBlock):
         """
         credit = self._determine_credit()
         self.score = credit.value
-        self.runtime.publish(
-            self,
-            'grade',
-            {
-                'value': self.score,
-                'max_value': Credit.full.value
-            }
-        )
+        try:
+            self.runtime.publish(
+                self,
+                'grade',
+                {
+                    'value': self.score,
+                    'max_value': Credit.full.value
+                }
+            )
+        except IntegrityError:
+            pass
 
     def _determine_credit(self):
         #  Not a standard xlbock pylint disable.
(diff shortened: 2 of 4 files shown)


freetextresponse/tests.py
```diff
@@ -260,6 +261,23 @@ def test_compute_score(self, credit):
             {'value': credit.value, 'max_value': Credit.full.value},
         )
 
+    @ddt.data(Credit.zero, Credit.half, Credit.full)
```
caseylitton: We should not need to run this test for each type of credit to verify the exception is handled properly.   We still need to mock determine_credit however, so just use any one of the credit values.

freetextresponse/tests.py
```diff
@@ -260,6 +261,23 @@ def test_compute_score(self, credit):
...
+    def test_compute_score_error(self, credit):
```
caseylitton: Error is pretty vague.  Let's make the function name more specific, like  test_compute_score_integrity_error

freetextresponse/tests.py
```diff
@@ -260,6 +261,23 @@ def test_compute_score(self, credit):
...
+        # pylint: disable=protected-access
+        """
+        Tests _compute_score's error handling
```
caseylitton: Same here too, let's use IntegrityError in the first line of the comment.  And be specific in the other parts of the comment as well. 

Does this test actually test that all exceptions are handle?  Or is it just the one IntegrityError?

caseylitton (changes requested):

caseylitton:
After you get two plus ones, you'll need to fix the commit message, 'Add try ... except'.  Your message needs to be less have 50 characters.  I assume that is where the '...' comes from

> @caseylitton The ellipsis may be coming from a lack of an empty line between the "subject" and "body"?

I didn't know that.  

@jlikhuva Make sure the commit message is less than 50 chars AND has an empty space after it if you add more comments after the message.  Fix this after the review is done.

freetextresponse/tests.py
jlikhuva: The current test only checks for the IntegrityError exception. The try - except that I introduced also only handles the IntegrityError exception. I could have it catch and ignore all exceptions, but I don't know if that is a good idea.

stvstnfrd: @jlikhuva Yeah, I think that's what @caseylitton is pointing out; the comment implies that this handles more than it actually does. We should only be catching that specific exception in this case; let's update the docs to reflect what we're catching.

stvstnfrd:
@caseylitton The ellipsis may be coming from a lack of an empty line between the "subject" and "body"?

freetextresponse/tests.py
caseylitton: Yes, thanks stv.  

@jlikhuva Specify exactly what's happening in the function.  We're forcing runtime.publish to throw an IntegrityError in order to test that compute_score passes it without issue.

jlikhuva:
@caseylitton @stvstnfrd  Fleshed out the documentation of the unit test. Also fixed the first commit message.

freetextresponse/tests.py
```diff
@@ -260,6 +261,25 @@ def test_compute_score(self, credit):
             {'value': credit.value, 'max_value': Credit.full.value},
         )
 
+    @ddt.data(Credit.zero)
```
caseylitton: Remove ddt entirely from this function.  We only need ddt when we want to run the same test with different parameters. 

In this case, we only need to use one credit value so...

return_value=credit   <---- Just set return_value to Credit.zero.  

Don't forget to remove credit from the function args.

caseylitton (changes requested):
Looks good.  Just remove unneeded ddt on your test.  

See if @caesar2164 can review before you push any more changes back up.  You can probably squash before the next push if he doesn't flag a major issue.

freetextresponse/tests.py
```diff
@@ -260,6 +261,25 @@ def test_compute_score(self, credit):
...
+    def test_compute_score_integrity_error(self, credit):
```
caesar2164: as @caseylitton said just remove the credit var being passed here.

freetextresponse/tests.py
```diff
@@ -260,6 +261,25 @@ def test_compute_score(self, credit):
...
+        is thrown by any of the methods/functions called in the
+        process of saving the score, the program handles it gracefully.
+        We force runtime.publish to throw an IntegrityError exception,
+        and expect _compute_score to be graceful about it.
+        """
+        self.xblock.runtime.publish = MagicMock(return_value=None)
+        self.xblock.runtime.publish.side_effect = IntegrityError(
+            "Unique Key Violation"
+        )
+        self.xblock._determine_credit = MagicMock(return_value=credit)
```
caesar2164: and just give this MagickMock `return_value=Credit.zero`

caesar2164 (changes requested):

jlikhuva:
@caesar2164 @caesar2164 

Fixed the ddt issue and Squashed commits.

caesar2164 (approved):
Perfect! 👍

caseylitton (approved):

caseylitton:
+1

stvstnfrd:
@jlikhuva Next time around, you'll want to clean up the commit message when squashing. Many of the old messages (e.g. `Fix Pylint error.`) are no longer relevant. Looking at the code as it exists on disk and the new, post-squash git commit history, there _technically_ never _was_ a Pylint error ;)
That's a development artifact that's we've (intentionally) erased from history/existence.
So you don't need to (and shouldn't) mention it in the final commit message.

PR #21, opened 2017-07-26
Move cached data from body to seq_content

potsui:
@stvstnfrd @caesar2164 PR

Previously, some future update or rogue script could easily tamper with
our cached data attributes on the body element. We moved these data
attributes to be slightly less widely accessible.

Arising from conversation in previous [PR](https://github.com/Stanford-Online/xblock-free-text-response/pull/18)
[Trello](https://trello.com/c/IHM9cBHS/953-move-cached-data-from-body-element-to-their-own-element-for-ftr-xblock)

CC: @jlikhuva @jspayd


caesar2164 (approved):
Looks good to me. (Granted I helped you write this... 😁 ) 👍

freetextresponse/private/view.js
```diff
@@ -3,6 +3,7 @@ function FreeTextResponseView(runtime, element) {
 
     var $ = window.jQuery;
     var $element = $(element);
+    var $seq_content = $('#seq_content');
```
stvstnfrd: We should give this a more generalized name.
One of the benefits to having a variable here is that we could update the selector if we want to store this elsewhere in the future.
In this case, if we updated the selector, the variable name would no longer make sense,
so we'd technically want to update it too.

Maybe `container` or something?
That's not a great name; any better ideas?

stvstnfrd (changes requested):
Code seems good, but we should chose a better variable name.

freetextresponse/private/view.js
potsui: `xblocksWrapper`? `contentContainer`? `xblocksContainer`? ¯\\\_(ツ)\_/¯

potsui:
On a related note, since this fix was originally taken from the submit and compare block, I've started a branch locally making the same fix for that xblock. Does that sound good?

freetextresponse/private/view.js
stvstnfrd: `xblocksContainer` seems reasonable.

stvstnfrd (approved):

stvstnfrd:
Go ahead and squash the commits and we'll merge then put this on stage tomorrow.

caesar2164 (changes requested):
Please bump version for this xblock, like this commit does: https://github.com/Stanford-Online/xblock-free-text-response/commit/ec5a21efd3357023b7b356cd1fefcccc3863d5e4

caesar2164 (approved):

PR #22, opened 2017-09-28
Add hint feedback

caseylitton:
Might be helpful when adding hint and feedback to free text?


stvstnfrd:
Closing for now;
branch archived here:
- https://github.com/Stanford-Online/xblock-free-text-response/tree/wip/caseylitton/add-hint-feedback

PR #24, opened 2017-11-13
Makes xblock due_date aware

caesar2164:
- Adds datetime and pytz packages
- Pushes version to 0.1.8

@dcadams & @stvstnfrd

### DIFF
--- a/freetextresponse/freetextresponse.py
+++ b/freetextresponse/freetextresponse.py
@@ -18,10 +18,11 @@ from xblock.fields import String
 from xblock.fragment import Fragment
 from xblock.validation import ValidationMessage
 from xblockutils.studio_editable import StudioEditableXBlockMixin
+from .mixins import EnforceDueDates
 
 
 @XBlock.needs("i18n")
-class FreeTextResponse(StudioEditableXBlockMixin, XBlock):
+class FreeTextResponse(EnforceDueDates, StudioEditableXBlockMixin, XBlock):
     #  pylint: disable=too-many-ancestors, too-many-instance-attributes
     """
     Enables instructors to create questions with free-text responses.
@@ -264,6 +265,7 @@ class FreeTextResponse(StudioEditableXBlockMixin, XBlock):
                 'problem_progress': self._get_problem_progress(),
                 'prompt': self.prompt,
                 'student_answer': self.student_answer,
+                'is_past_due': self.is_past_due(),
                 'used_attempts_feedback': self._get_used_attempts_feedback(),
                 'visibility_class': self._get_indicator_visibility_class(),
                 'word_count_message': self._get_word_count_message(),
@@ -539,6 +541,15 @@ class FreeTextResponse(StudioEditableXBlockMixin, XBlock):
             result = self._get_invalid_word_count_message(ignore_attempts)
         return result
 
+    def _can_submit(self):
+        if self.is_past_due():
+            return False
+        if self.max_attempts == 0:
+            return True
+        if self.count_attempts < self.max_attempts:
+            return True
+        return False
+
     @XBlock.json_handler
     def submit(self, data, suffix=''):
         # pylint: disable=unused-argument
@@ -547,7 +558,7 @@ class FreeTextResponse(StudioEditableXBlockMixin, XBlock):
         """
         # Fails if the UI submit/save buttons were shut
         # down on the previous sumbisson
-        if self.max_attempts == 0 or self.count_attempts < self.max_attempts:
+        if self._can_submit():
             self.student_answer = data['student_answer']
             # Counting the attempts and publishing a score
             # even if word count is invalid.
--- a/freetextresponse/mixins.py
+++ b/freetextresponse/mixins.py
@@ -1,0 +1,33 @@
+"""
+Mixins for the Free Text Response XBlock
+"""
+import datetime
+import pytz
+from django.conf import settings
+
+TIME_ZONE = pytz.timezone(getattr(settings, 'TIME_ZONE', pytz.utc.zone))
+
+
+class EnforceDueDates(object):  # pylint: disable=too-few-public-methods
+    """
+    xBlock Mixin to allow xblocks to check the due date
+    (taking the graceperiod into account) of the
+    subsection in which they are placed
+    """
+
+    # These values are pulled from platform.
+    # They are defaulted to None for tests.
+    due = None
+    graceperiod = None
+
+    def is_past_due(self):
+        """
+        Determine if component is past-due
+        """
+        now = datetime.datetime.utcnow().replace(tzinfo=TIME_ZONE)
+        if self.due is not None:
+            due_date = self.due
+            if self.graceperiod is not None:
+                due_date = due_date + self.graceperiod
+            return now > due_date
+        return False
--- a/freetextresponse/templates/freetextresponse_view.html
+++ b/freetextresponse/templates/freetextresponse_view.html
@@ -12,12 +12,14 @@
     </div>
     <div class="capa_alert submission-received">{{ submitted_message }}</div>
     <div class="action">
-        <button class="check Submit {{ nodisplay_class }}" data-checking="{% trans "Checking..." %}" data-value={% trans "Submit" %}>
-            {% trans "Submit" %}
-        </button>
-        <button class="save {{ nodisplay_class }}" data-checking="{% trans "Checking..." %}" data-value={% trans "Save" %}>
-            {% trans "Save" %}
-        </button>
+        {% if not is_past_due %}
+            <button class="check Submit {{ nodisplay_class }}" data-checking="{% trans "Checking..." %}" data-value={% trans "Submit" %}>
+                {% trans "Submit" %}
+            </button>
+            <button class="save {{ nodisplay_class }}" data-checking="{% trans "Checking..." %}" data-value={% trans "Save" %}>
+                {% trans "Save" %}
+            </button>
+        {% endif %}
         <div class="used-attempts-feedback" aria-live="polite">{{ used_attempts_feedback }}</div>
     </div>
     <div class="capa_alert user_alert">{{ user_alert }}</div>
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "xblock-free-text-response",
   "title": "FreeTextResponse XBlock",
   "description": "Enables instructors to create questions with free-text responses.",
-  "version": "0.1.7",
+  "version": "0.1.8",
   "homepage": "https://github.com/Stanford-Online/xblock-free-text-response",
   "author": {
     "name": "Azim Pradhan",
--- a/setup.py
+++ b/setup.py
@@ -27,7 +27,7 @@ class Tox(TestCommand):
 
 setup(
     name="xblock-free-text-response",
-    version="0.1.7",
+    version="0.1.8",
     description="Enables instructors to create questions with free-text responses.",
     license='AGPL-3.0',
     packages=[


freetextresponse/mixins.py
```diff
@@ -0,0 +1,39 @@
...
+Due Date Integration xBlock Mixin
+"""
+import datetime
+import pytz
+from django.conf import settings
+from xblock.fields import DateTime
+from xblock.fields import Scope
+
+
+class DueDateXblockMixin(object):
```
stvstnfrd: I know edx does it everywhere, but I'm not a fan of repeating "obvious" words over and over.
In this case, we're in an XBlock, writing code for the XBlock Utils package.
It should be obvious enough that it's related to XBlocks,
without including that in the class name. Similarly with `mixin`

For mixins/interfaces, since they're usually describing an ability or behavior,
I like to use that to describe the name.
Here, since we're talking about enforcing due dates, that could be something like:

`class EnforceDueDates(object):`

That then gives us something like:
`class MyXBlock(EnforceDueDates, ...):`
which makes it more clear that "oh, my xblock enforces due dates" :)

Either way, any time we write XBlock like this, we should do it consistently (`XBlock` vs `Xblock`)

freetextresponse/mixins.py
```diff
@@ -0,0 +1,39 @@
+"""
+Due Date Integration xBlock Mixin
```
stvstnfrd: This docstring is supposed to be about the file, not the class.
You could have additional mixins here.

freetextresponse/mixins.py
```diff
@@ -0,0 +1,39 @@
...
+from xblock.fields import DateTime
+from xblock.fields import Scope
+
+
+class DueDateXblockMixin(object):
+    """
+    xBlock Mixin to allow xblocks to check the due date of
+    the subsection in which they are placed
+    """
+    due = DateTime(
```
stvstnfrd: You said we didn't need this;
what changed?

freetextresponse/mixins.py
```diff
@@ -0,0 +1,39 @@
...
+        scope=Scope.settings,
+        default=None,
+    )
+    graceperiod = None
```
stvstnfrd: What's this?

Specifically, where did you get it from and why do we need it?

freetextresponse/mixins.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def past_due(self):
```
stvstnfrd: Booleans are more readable when prefixed by `is/should/has/etc..`:

```python
if not is_past_due():
    # allow submission
```

This also makes the type more obvious,
which is important here since `due` looks similar (`due` vs `past_due`),
but one is actually a `DateTime`, while this one is a `boolean`.

freetextresponse/mixins.py
```diff
@@ -0,0 +1,39 @@
...
+        """
+        Return whether due date has passed.
```
stvstnfrd: `Return` is almost always superfluous in a docstring description,
since all functions return some value.

```
Determine if component is past-due

This can be used to determine whether users should be allow to submit responses.
```

You _could_ even drop the `Determine` and leave it at
`Is component past-due`, though leading verbs do lead to readability.

freetextresponse/mixins.py
```diff
@@ -0,0 +1,39 @@
...
+        """
+        site_tz = pytz.timezone(getattr(settings, 'TIME_ZONE', pytz.utc.zone))
```
stvstnfrd: Since `settings` is static at run-time,
we don't need to recalculate it on each page load.

You can memoize this into a constant at the top-level.

freetextresponse/mixins.py
```diff
@@ -0,0 +1,39 @@
...
+        now = datetime.datetime.utcnow().replace(tzinfo=site_tz)
+        if self.due is not None:
+            due_date = self.due
+            if self.graceperiod is not None:
```
stvstnfrd: How is `graceperiod` ever anything but `None`?

freetextresponse/mixins.py
```diff
@@ -0,0 +1,39 @@
...
+                due_date = due_date + self.graceperiod
+            return now > due_date
+        return False
+
+    def before_due(self):
```
stvstnfrd: I'm not a fan of this composite value.

"past due" is an expected description of an assignment,
not just in code but also IRL.
It's unlikely that anyone would describe an assignment as "before due".

Also, it's one more function to remember/test
and worse, one more thing that can go wrong.

As seen here:
```python
past_due = now > due_date
before_due = not past_due
# which can be rewritten as
before_due = not (now > due_date)
before_due = now <= due_date
# which means that
before_due is True  # when due_date == now
# which is misleading
```

It becomes easier to miss these edge cases, the more composite values are used.

We should remove this method and just say:
` if not past_due`, wherever necessary.

freetextresponse/mixins.py
```diff
@@ -0,0 +1,39 @@
...
+import pytz
+from django.conf import settings
+from xblock.fields import DateTime
+from xblock.fields import Scope
+
+
+class DueDateXblockMixin(object):
+    """
+    xBlock Mixin to allow xblocks to check the due date of
+    the subsection in which they are placed
```
stvstnfrd: Can they have their own due date, different from the subsection?

freetextresponse/mixins.py
stvstnfrd: Trivia: by convention, interfaces (in languages like C#, Java; _not_ Python) often are prefixed with a lowercase `i` (for interface),
which gives you names like `iEnforceDueDates`
which I always like reading as a boastful first-person declaration:

> What do you do?

> Me? I enforce due dates

:)

stvstnfrd (changes requested):

freetextresponse/mixins.py
```diff
@@ -0,0 +1,34 @@
...
+    xBlock Mixin to allow xblocks to check the due date
+    (taking the graceperiod into account) of the
+    subsection in which they are placed
+    """
+    # get instance timezone
+    site_tz = pytz.timezone(getattr(settings, 'TIME_ZONE', pytz.utc.zone))
+
+    # Default values pulled from platform to None for tests
+    due = None
+    graceperiod = None
```
caesar2164: This pulls value from Grading page in studio (matches CAPA)

freetextresponse/mixins.py
```diff
@@ -0,0 +1,34 @@
...
+    """
+    xBlock Mixin to allow xblocks to check the due date
+    (taking the graceperiod into account) of the
+    subsection in which they are placed
+    """
+    # get instance timezone
+    site_tz = pytz.timezone(getattr(settings, 'TIME_ZONE', pytz.utc.zone))
+
+    # Default values pulled from platform to None for tests
+    due = None
```
caesar2164: Needed for tests, gets overridden with subsection value.

freetextresponse/mixins.py
```diff
@@ -0,0 +1,34 @@
...
+from django.conf import settings
+from xblock.fields import DateTime
+from xblock.fields import Scope
+
+
+class EnforceDueDates(object):
+    """
+    xBlock Mixin to allow xblocks to check the due date
+    (taking the graceperiod into account) of the
+    subsection in which they are placed
```
caesar2164: Xblocks *could* set further due dates but different logic would have to be written in the `_can_submit()`

freetextresponse/mixins.py
```diff
@@ -0,0 +1,32 @@
...
+    """
+    # get instance timezone
+    self.site_tz = pytz.timezone(getattr(settings, 'TIME_ZONE', pytz.utc.zone))
+
+    # Default values pulled from platform to None for tests
+    due = None
```
caesar2164: Needed for tests, gets overridden with subsection value.

freetextresponse/mixins.py
```diff
@@ -0,0 +1,32 @@
...
+    graceperiod = None
```
caesar2164: This pulls value from Grading page in studio (matches CAPA). Needs to be set here to `None` for tests.

freetextresponse/mixins.py
```diff
@@ -0,0 +1,32 @@
+"""
+Mixins for the Free Text Response XBlock
+"""
+import datetime
+import pytz
+from django.conf import settings
+
+
+class EnforceDueDates(object):
```
caesar2164: Pylint is breaking here.
It's set to require 2 public methods per class.

I basically had added that `before_due()` function so it would have 2 public methods

stvstnfrd: @caesar2164 Can you paste the actual error message/code?

caesar2164:
```
R:  9, 0: Too few public methods (1/2) (too-few-public-methods)
```

caesar2164: I could add a second method?
```python
def has_graceperiod(self):
    """
    Determine if course has a graceperiod set
    """
    if self.graceperiod is not None:
        return True
    return False
```

stvstnfrd: No, we should never write unused code, just to satisfy static analysis tools.

People commonly miss the _spirit_ of the rule,
in favor of the _letter_ of the law.

Pylint errors don't mean:

> This is never acceptable code; change it immediately!

It's more nuanced; what they're really saying is:

> While syntactically valid, this design pattern is _often_ indicative of a mistake/oversight and/or can _often_ lead to issues later on.

Specifically here, what they're saying is:

> Generally speaking, classes should have more than one method. If your implementation doesn't, then maybe a dedicated `class` is overkill and a lighter abstraction is more appropriate.

In this case, our response is along the lines of:

> Yeah, we know it's lightweight. It's actually not meant to be a standalone class, but rather mixed-in to base class. We appreciate your concern, but it's nothing to worry about.

And we "respond" to that by ignoring the pylint error.

Similarly, in an ideal world, ignoring pylint doesn't mean:

> I'm too lazy to fix this

Instead it _should_ mean:

> WONTFIX: for reasons

freetextresponse/freetextresponse.py
```diff
@@ -264,6 +265,7 @@ def student_view(self, context={}):
                 'problem_progress': self._get_problem_progress(),
                 'prompt': self.prompt,
                 'student_answer': self.student_answer,
+                'before_due': not self.is_past_due(),
```
stvstnfrd: I'd prefer to use the term `past_due` consistently throughout.

freetextresponse/mixins.py
```diff
@@ -0,0 +1,32 @@
...
+    """
+    xBlock Mixin to allow xblocks to check the due date
+    (taking the graceperiod into account) of the
+    subsection in which they are placed
+    """
+    # get instance timezone
+    site_tz = pytz.timezone(getattr(settings, 'TIME_ZONE', pytz.utc.zone))
```
stvstnfrd: Please move this to the top-level of the file;
it doesn't need recalculated for each instantiation of the class
(which is basically every page load).

```python
TIME_ZONE = pytz.timezone(getattr(settings, 'TIME_ZONE', pytz.utc.zone))
```

freetextresponse/freetextresponse.py
caesar2164: I actually specifically made this `before_due` to make the django template more readable:

```django
{% if before_due %}
    # Stuff
{% endif %}
```

vs

```django
{% if past_due != None%}
    # Stuff
{% endif %}
```

freetextresponse/mixins.py
```diff
@@ -0,0 +1,33 @@
+"""
+Mixins for the Free Text Response XBlock
+"""
+import datetime
+import pytz
+from django.conf import settings
+
+# get instance timezone
```
stvstnfrd: I think this code is clear enough without the comment.

freetextresponse/freetextresponse.py
```diff
@@ -264,6 +265,7 @@ def student_view(self, context={}):
                 'problem_progress': self._get_problem_progress(),
                 'prompt': self.prompt,
                 'student_answer': self.student_answer,
+                'past_due': self.is_past_due(),
```
stvstnfrd: Kind of a nit, but why not just use `is_past_due` everywhere?

freetextresponse/freetextresponse.py
```diff
@@ -539,6 +541,15 @@ def _get_user_alert(self, ignore_attempts=False):
             result = self._get_invalid_word_count_message(ignore_attempts)
         return result
 
+    def _can_submit(self):
+        if not self.is_past_due():
```
stvstnfrd: This isn't correct; I'll come explain f2f.

freetextresponse/mixins.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+class EnforceDueDates(object):  # pylint: disable=too-few-public-methods
+    """
+    xBlock Mixin to allow xblocks to check the due date
+    (taking the graceperiod into account) of the
+    subsection in which they are placed
+    """
+
+    # Default values pulled from platform to None for tests
```
stvstnfrd: I think you misworded this, specifically `pulled from platform _to_ None for tests`

stvstnfrd (approved):

PR #25, opened 2017-11-13
Check due dates via mixin

stvstnfrd:
@caesar2164 I don't have a workbench ready to test this in, but this is what the mixin looks like.


stvstnfrd:
@caesar2164 Do we still need this for anything?

caesar2164:
nope, @stvstnfrd

PR #26, opened 2017-11-16
Fix is_past_due function

caesar2164:
@stvstnfrd


stvstnfrd:
+1, but check w/ @dcadams before merging.

dcadams:
:+1:

PR #30, opened 2018-07-26
Add option to display other student responses.

saharm1:
@stvstnfrd  @caesar2164 

lots of files changed because of grunt, including last package file.


freetextresponse/settings.py
```diff
@@ -16,4 +16,6 @@
     'freetextresponse',
 )
 
+ALLOWED_HOSTS = ['*']
```
saharm1: According to stackoverflow, need to add this field, which solves the issue. But * should also not be included in production, so I need to figure out what the allowed hosts actually are.

stvstnfrd: > According to stackoverflow, need to add this field, which solves the issue

@saharm1 I don't think I'm clear on the what issue this aims to solve...

freetextresponse/freetextresponse.py
```diff
@@ -18,11 +18,69 @@
 from xblock.fragment import Fragment
 from xblock.validation import ValidationMessage
 from xblockutils.studio_editable import StudioEditableXBlockMixin
+from xblockutils.publish_event import PublishEventMixin
```
saharm1: need to remove

freetextresponse/freetextresponse.py
```diff
@@ -18,11 +18,69 @@
...
 from .mixins import EnforceDueDates
 
+from answer_pool import offer_answer, get_other_answers
+
+class MissingDataFetcherMixin:
```
saharm1: copied from peer submissions, who copied from elsewhere

freetextresponse/freetextresponse.py
caesar2164: remember to remove all the instances of use below in this file.

freetextresponse/templates/freetextresponse_view.html
```diff
@@ -23,4 +25,19 @@ <h2 class="problem-header">{{ display_name }}</h2>
         <div class="used-attempts-feedback" aria-live="polite">{{ used_attempts_feedback }}</div>
     </div>
     <div class="capa_alert user_alert">{{ user_alert }}</div>
+    <ul class="responses-box" id="responses-box" style="display:none;">
+        <button class="hide" id="hide-button" onclick="hideResponses()"> Hide </button>
+        <p class="responses-title"> Submissions by others </p>
+        <li class="response-list" id="response-list"</li>
```
caesar2164: opening `<li>` tag is missing the closing carrot `>`

freetextresponse/answer_pool.py
```diff
@@ -0,0 +1,49 @@
...
+        'student_id': student_id,
+        'answer': answer,
+    })
+    while sys.getsizeof(pool) > POOL_SIZE:
+        del pool[0]
+
+def get_other_answers(pool, get_student_item_dict):
+    ret = []
+    student_id = get_student_item_dict['student_id']
+    ret_length = min(3, len(pool))
```
caesar2164: make the number a Constant!

freetextresponse/answer_pool.py
```diff
@@ -0,0 +1,49 @@
...
+                ...
+            ]
+        answer: the answer the student gave
+        student_id: student identifier
+    """
+    pool.append({
+        'student_id': student_id,
+        'answer': answer,
+    })
+    while sys.getsizeof(pool) > POOL_SIZE:
```
caesar2164: Add comment to explain what this is doing (maybe just add it to the docstring?)

freetextresponse/templates/freetextresponse_view.html
```diff
@@ -10,6 +10,8 @@ <h2 class="problem-header">{{ display_name }}</h2>
             <span class="status {{ visibility_class }}" aria-describedby="student_answer"></span>
         </div>
     </div>
+    <input id="messageCheckbox" type="checkbox"> Allow my response
```
caesar2164: make this default to checked

freetextresponse/templates/freetextresponse_view.html
```diff
@@ -23,4 +25,19 @@ <h2 class="problem-header">{{ display_name }}</h2>
...
     </div>
     <div class="capa_alert user_alert">{{ user_alert }}</div>
+    <ul class="responses-box" id="responses-box" style="display:none;">
+        <button class="hide" id="hide-button" onclick="hideResponses()"> Hide </button>
+        <p class="responses-title"> Submissions by others </p>
+        <li class="response-list" id="response-list"</li>
+    </ul>
 </div>
+<script>
+function hideResponses() {
```
caesar2164: let's move this to the JS

freetextresponse/public/view.js
```diff
@@ -74,6 +79,18 @@ function FreeTextResponseView(runtime, element) {
...
     });
 
+    function htmlBuilder(responses) {
+        var html = '';
+        responses.forEach(function(item) {
+            html = html + '<div id="other-student-responses">' +
+            item.answer + '</div>';
+        });
+        if (html === '') {
+            html = '<p id="no-response"> No responses to show at this time </p>';
```
caesar2164: period at end of sentence

freetextresponse/public/view.js
```diff
@@ -57,12 +58,16 @@ function FreeTextResponseView(runtime, element) {
                 buttonSubmit.text(buttonSubmit[0].dataset.value);
                 userAlertMessage.text(response.user_alert);
                 buttonSave.addClass(response.nodisplay_class);
-                setClassForTextAreaParent(response.indicator_class); 
+                setClassForTextAreaParent(response.indicator_class);
+                if (!response.user_alert && response.display_other_responses) {
+                    responseList.html(htmlBuilder(response.other_responses));
+                    document.getElementById("responses-box").style.display = "block";
```
caesar2164: let's make a helper function to show and hide the responses box

freetextresponse/public/view.js
```diff
@@ -47,7 +47,8 @@ function FreeTextResponseView(runtime, element) {
         $.ajax(url, {
             type: 'POST',
             data: JSON.stringify({
-                'student_answer': $element.find('.student_answer').val()
+                'student_answer': $element.find('.student_answer').val(),
+                'checked': document.getElementById('messageCheckbox').checked
```
caesar2164: could you use `$element.find()` for this?

caesar2164 (changes requested):
First round of review

freetextresponse/templates/freetextresponse_view.html
```diff
@@ -23,4 +25,19 @@ <h2 class="problem-header">{{ display_name }}</h2>
...
+    var x = document.getElementById("response-list");
+    if (x.style.display === "none") {
```
stvstnfrd: jQuery actually has a builtin `toggle` method:
https://api.jquery.com/toggle/

freetextresponse/templates/freetextresponse_view.html
```diff
@@ -23,4 +25,19 @@ <h2 class="problem-header">{{ display_name }}</h2>
         <div class="used-attempts-feedback" aria-live="polite">{{ used_attempts_feedback }}</div>
     </div>
     <div class="capa_alert user_alert">{{ user_alert }}</div>
+    <ul class="responses-box" id="responses-box" style="display:none;">
+        <button class="hide" id="hide-button" onclick="hideResponses()"> Hide </button>
+        <p class="responses-title"> Submissions by others </p>
```
stvstnfrd: Don't wrap text w/ surrounding spaces. Start/end immediately after/before the angle brackets:

```html
<p>foobar</p>
```

freetextresponse/templates/freetextresponse_view.html
```diff
@@ -23,4 +25,19 @@ <h2 class="problem-header">{{ display_name }}</h2>
         <div class="used-attempts-feedback" aria-live="polite">{{ used_attempts_feedback }}</div>
     </div>
     <div class="capa_alert user_alert">{{ user_alert }}</div>
+    <ul class="responses-box" id="responses-box" style="display:none;">
```
stvstnfrd: Move all styling to CSS, instead of inline styles.

stvstnfrd: You can't use ids in an xblock.
Really, you should avoid using IDs in HTML, period, as a general rule..
IDs must only appear on a page at _most_ once,
but an instructor should be able to add multiple instances of this xblock to the same page.

Instead of IDs here, you'll need to use a combination of classes and parent selectors, see:
https://github.com/Stanford-Online/xblock-free-text-response/blob/master/freetextresponse/private/view.js#L9-L13

freetextresponse/templates/freetextresponse_view.html
```diff
@@ -23,4 +25,19 @@ <h2 class="problem-header">{{ display_name }}</h2>
...
+        <button class="hide" id="hide-button" onclick="hideResponses()"> Hide </button>
```
stvstnfrd: Don't inline click handler; bind in JS. See:
https://github.com/Stanford-Online/xblock-free-text-response/blob/master/freetextresponse/private/view.js#L41

freetextresponse/public/view.js
```diff
@@ -74,6 +79,18 @@ function FreeTextResponseView(runtime, element) {
         return false;
     });
 
+    function htmlBuilder(responses) {
```
stvstnfrd: TODO: comment more here...

freetextresponse/public/view.js
stvstnfrd: @caesar2164 jQuery has builtin helpers for `show`, `hide`, and `toggle` ;)

freetextresponse/answer_pool.py
```diff
@@ -0,0 +1,49 @@
+import random
+import sys
+import persistence as sas_api
+from utils import _ #pylint: disable=unused-import
+
+# pool size in bytes, the whole pool will be stored in the block and retrieved
+# each time the page loads so we want to keep the pool to a reasonable small size.
+POOL_SIZE = 4096
```
stvstnfrd: You cribbed this from somewhere, right?
We should link to the related code..

freetextresponse/answer_pool.py
```diff
@@ -0,0 +1,49 @@
...
+    """
+    pool.append({
+        'student_id': student_id,
+        'answer': answer,
+    })
+    while sys.getsizeof(pool) > POOL_SIZE:
+        del pool[0]
+
+def get_other_answers(pool, get_student_item_dict):
+    ret = []
```
stvstnfrd: Use a better variable name.

freetextresponse/answer_pool.py
```diff
@@ -0,0 +1,49 @@
+import random
+import sys
+import persistence as sas_api
+from utils import _ #pylint: disable=unused-import
```
stvstnfrd: Are we using this? If not, remove.

freetextresponse/answer_pool.py
```diff
@@ -0,0 +1,49 @@
+import random
+import sys
+import persistence as sas_api
```
stvstnfrd: Are we using this? If not, remove.

freetextresponse/answer_pool.py
```diff
@@ -0,0 +1,49 @@
+import random
```
stvstnfrd: Are we using this? If not, remove.

freetextresponse/freetextresponse.py
```diff
@@ -18,11 +18,69 @@
 from xblock.fragment import Fragment
 from xblock.validation import ValidationMessage
 from xblockutils.studio_editable import StudioEditableXBlockMixin
+from xblockutils.publish_event import PublishEventMixin
 from .mixins import EnforceDueDates
 
+from answer_pool import offer_answer, get_other_answers
```
stvstnfrd: This is a local import, right?
If so, it should be moved up a line (to be grouped with `.mixins` above).
It should also be a relative import (again, like mixins above).

freetextresponse/freetextresponse.py
stvstnfrd: We're writing python2, so you should use "new-style" classes that inherit from `object`.

https://stackoverflow.com/a/45062077

CC: @Stanford-Online/curis

freetextresponse/public/view.js
stvstnfrd: Yes, it's used elsewhere in this xblock..

freetextresponse/public/view.less
```diff
@@ -111,5 +111,48 @@
...
+  }
+
+  #hide-button {
+    float: right;
+    background:none;
+    border:none;
+    padding:0;
+    color:#069;
+    text-decoration:underline;
+    cursor:pointer;
```
stvstnfrd: always put a space after the colon `:`.

freetextresponse/public/view.js
stvstnfrd: @caesar2164 It isn't a sentence though...
Should it be?

freetextresponse/public/view.js
```diff
@@ -74,6 +79,18 @@ function FreeTextResponseView(runtime, element) {
...
+        var html = '';
+        responses.forEach(function(item) {
+            html = html + '<div id="other-student-responses">' +
```
stvstnfrd: I think you can put this on a single line like:

```js
html += '<div id="other-student-responses">' + item.answer + '</div>';
```

freetextresponse/freetextresponse.py
```diff
@@ -599,6 +683,15 @@ def save_reponse(self, data, suffix=''):
...
 
+    def handle_response(self, answer, record_response=False):
+        """
+        Store response from students to the backend
+        """
+        student_item = self.get_student_item_dict()
+        if record_response:
+            offer_answer(self.system_answers, answer, student_item['student_id'])
+        self.other_responses_shown = get_other_answers(
+            self.system_answers, student_item)
```
stvstnfrd: Either place this all on a single line or split the two args:

```py
self.other_responses_shown = get_other_answers(self.system_answers, student_item)
self.other_responses_shown = get_other_answers(
    self.system_answers,
    student_item,
)
```

freetextresponse/freetextresponse.py
```diff
@@ -599,6 +683,15 @@ def save_reponse(self, data, suffix=''):
...
         return result
 
+    def handle_response(self, answer, record_response=False):
+        """
+        Store response from students to the backend
+        """
+        student_item = self.get_student_item_dict()
+        if record_response:
+            offer_answer(self.system_answers, answer, student_item['student_id'])
+        self.other_responses_shown = get_other_answers(
```
stvstnfrd: Wait, do we actually need to update the `other_responses_shown` list if we aren't "offering" a new response? Shouldn't the list be unchanged?

freetextresponse/freetextresponse.py
```diff
@@ -564,6 +643,9 @@ def submit(self, data, suffix=''):
             # even if word count is invalid.
             self.count_attempts += 1
             self._compute_score()
+            #check if valid
```
stvstnfrd: Make sure you run pep8 and pylint against this code..

freetextresponse/freetextresponse.py
```diff
@@ -599,6 +683,15 @@ def save_reponse(self, data, suffix=''):
         }
         return result
 
+    def handle_response(self, answer, record_response=False):
```
stvstnfrd: Would a more apt name be something like `store_student_response`?

freetextresponse/freetextresponse.py
```diff
@@ -86,6 +144,23 @@ def workbench_scenarios():
...
+        ),
+        default=False,
+        scope=Scope.settings,
+    )
+    system_answers = List(
+        default=[], scope=Scope.user_state_summary,
+        help=_("System selected answers to give to students"),
+    )
+    other_responses_shown = List(
+        default=[], scope=Scope.user_state,
```
stvstnfrd: Per style guides, if you're splitting _any_ kwargs across multiple lines, you have to split _all_ of them.

freetextresponse/freetextresponse.py
```diff
@@ -18,11 +18,69 @@
...
+                student_id = anonymous_user_id
+            else:
+                student_id = self.xmodule_runtime.anonymous_student_id  #pylint:disable=E1101
+        else:
+            course_id = "edX/Enchantment_101/April_1"
+            if self.scope_ids.user_id is None:
+                student_id = ''
+            else:
+                student_id = unicode(self.scope_ids.user_id)
+        student_item_dict = dict(
```
stvstnfrd: Typically, if we're creating a literal dictionary in python, just use the literal syntax, instead of the `dict` constructor:

```py
student_item_dict = {
    'student_id': student_id,
    # ...
}
```

freetextresponse/freetextresponse.py
```diff
@@ -18,11 +18,69 @@
...
+        item_id = self._serialize_opaque_key(self.scope_ids.usage_id)
+
+        if hasattr(self, "xmodule_runtime"):
+            course_id = self.get_course_id()  # pylint:disable=E1101
+            if anonymous_user_id:
+                student_id = anonymous_user_id
+            else:
+                student_id = self.xmodule_runtime.anonymous_student_id  #pylint:disable=E1101
+        else:
+            course_id = "edX/Enchantment_101/April_1"
```
stvstnfrd: ?!

freetextresponse/freetextresponse.py
```diff
@@ -18,11 +18,69 @@
...
+            if self.scope_ids.user_id is None:
```
stvstnfrd: I think you could simplify this like:

```py
student_id = self.scope_ids.user_id or ''
student_id = unicode(student_id)
```

since both `0` and the empty string `''` are invalid student ids.

stvstnfrd (changes requested):
@saharm1 Here's a quick once-over. I'll take a deeper look later.

freetextresponse/freetextresponse.py
saharm1: This is super interesting, I had no idea!

freetextresponse/public/view.js
saharm1: I am having a problem using show or toggle here, since it reverts to previous display type. The initial type is none, so it reverts to no display.

freetextresponse/freetextresponse.py
saharm1: I believe this is the first time this value is set.

freetextresponse/public/view.js
caesar2164: `.show()` should do exactly what you're doing here manually...

freetextresponse/answer_pool.py
```diff
@@ -0,0 +1,59 @@
...
+        'student_id': student_id,
+        'answer': answer,
+    })
+    while sys.getsizeof(pool) > POOL_SIZE:
+        del pool[0]
+
+
+def get_other_answers(pool, get_student_item_dict):
+    """
+    Returns at most the three latest answers from the pool.
```
caesar2164: rather than "three" maybe let's say: `Returns up to the preset number of answers...`

freetextresponse/answer_pool.py
```diff
@@ -0,0 +1,59 @@
...
+    Does not return answers the student had submitted.
+    """
+    return_list = []
+    student_id = get_student_item_dict['student_id']
+    return_list_length = min(MAX_RESPONSES, len(pool))
+
+    for answer in reversed(pool):
+        if len(return_list) >= return_list_length:
```
caesar2164: Why does `return_list_length` need to be calculated? Couldn't you just say:
```python
for answer in reversed(pool):
    if len(return_list) >= MAX_RESPONSES:
        break
    ...
```

freetextresponse/freetextresponse.py
```diff
@@ -18,11 +18,80 @@
...
+
+        if hasattr(self, "xmodule_runtime"):
+            course_id = self.get_course_id()  # pylint:disable=E1101
+            if anonymous_user_id:
+                student_id = anonymous_user_id
+            else:
+                student_id = self.xmodule_runtime.anonymous_student_id
+                # pylint:disable=E1101
+        else:
+            course_id = "edX/Enchantment_101/April_1"
```
caesar2164: What's this weird course ID? Also why is it the default for this?

freetextresponse/freetextresponse.py
```diff
@@ -269,8 +361,12 @@ def student_view(self, context={}):
                 'used_attempts_feedback': self._get_used_attempts_feedback(),
                 'visibility_class': self._get_indicator_visibility_class(),
                 'word_count_message': self._get_word_count_message(),
+                'display_other_responses':
+                    self.display_other_student_responses,
+                'other_responses': self.other_responses_shown,
             }
         )
+
```
caesar2164: remove added empty line

caesar2164 (changes requested):
Round 2!

freetextresponse/templates/freetextresponse_view.html
```diff
@@ -23,4 +25,9 @@ <h2 class="problem-header">{{ display_name }}</h2>
         <div class="used-attempts-feedback" aria-live="polite">{{ used_attempts_feedback }}</div>
     </div>
     <div class="capa_alert user_alert">{{ user_alert }}</div>
+    <ul class="responses-box" id="responses-box">
+        <button class="hide-button">Hide</button>
+        <p class="responses-title">Submissions by others</p>
+        <li class="response-list"></li>
```
caesar2164: Why are we using `<ul>` and `<li>` elements for displaying responses?

Especially since the "list" actually seems to be contained within a list item?

freetextresponse/public/view.less
```diff
@@ -111,5 +111,50 @@
...
+  }
+
+  .response-list {
+    list-style: none;
+    text-align: center;
+  }
+
+  .other-student-responses {
+    display: inline-block;
+    background: white;
```
caesar2164: do we NEED to set a background color? if we don't have to, it's better not to, so we avoid the white on off-white, like in the forum...

freetextresponse/public/view.less
```diff
@@ -111,5 +111,50 @@
...
+    width: 97%;
```
caesar2164: What's the rationale behind 97%?

freetextresponse/public/view.js
```diff
@@ -74,6 +84,22 @@ function FreeTextResponseView(runtime, element) {
...
+        Convert list of responses to a html string to add to the page
+        */
+        var html = '';
+        //for each item in the list, add a div containing it to the html
+        responses.forEach(function(item) {
+            html += '<div class="other-student-responses">' + item.answer + '</div>';
+        });
+        //if the answer list is empty, add the appropriate string to the page
+        if (html === '') {
+            html = '<p class="no-response"> No responses to show at this time </p>';
```
caesar2164: if you default `response-list` to have this HTML in it, and replace only when there is content you can simplify this a lot.

caesar2164 (changes requested):

freetextresponse/templates/freetextresponse_view.html
```diff
@@ -23,4 +25,9 @@ <h2 class="problem-header">{{ display_name }}</h2>
         <div class="used-attempts-feedback" aria-live="polite">{{ used_attempts_feedback }}</div>
     </div>
     <div class="capa_alert user_alert">{{ user_alert }}</div>
+    <ul class="responses-box" id="responses-box">
```
saharm1: responses-box = div, response-list = ul, reponses = li

freetextresponse/private/view.less
```diff
@@ -111,5 +111,50 @@
     font-size: 0.9em;
   }
 
+  .responses-box {
+    display: none;
+    margin-top: 10px;
+    background: #f8f8f8;
```
saharm1: fbfbfb

freetextresponse/private/view.less
```diff
@@ -111,5 +111,50 @@
...
+  }
+
+  .response-list {
+    list-style: none;
+    text-align: center;
+  }
+
+  .other-student-responses {
+    display: inline-block;
+    background: white;
```
saharm1: dont define, put border (1 pixel dotted or dashed, light, same color as main border)

freetextresponse/public/view.less
saharm1: width: 100%, box-sizing: border-box;

freetextresponse/templates/freetextresponse_view.html
saharm1: automatic no responses to display

freetextresponse/public/view.js
```diff
@@ -74,6 +84,22 @@ function FreeTextResponseView(runtime, element) {
...
+    function htmlBuilder(responses) {
+        /*
+        Convert list of responses to a html string to add to the page
+        */
+        var html = '';
+        //for each item in the list, add a div containing it to the html
+        responses.forEach(function(item) {
+            html += '<div class="other-student-responses">' + item.answer + '</div>';
+        });
+        //if the answer list is empty, add the appropriate string to the page
```
saharm1: remove after html default

freetextresponse/public/view.js
```diff
@@ -57,12 +63,16 @@ function FreeTextResponseView(runtime, element) {
                 buttonSubmit.text(buttonSubmit[0].dataset.value);
                 userAlertMessage.text(response.user_alert);
                 buttonSave.addClass(response.nodisplay_class);
-                setClassForTextAreaParent(response.indicator_class); 
+                setClassForTextAreaParent(response.indicator_class);
+                if (!response.user_alert && response.display_other_responses) {
+                    responseList.html(htmlBuilder(response.other_responses));
```
saharm1: var responsehtml = builder, if responsehtml then edit html

freetextresponse/private/view.less
```diff
@@ -111,5 +111,50 @@
...
+  .responses-box {
+    display: none;
+    margin-top: 10px;
+    background: #f8f8f8;
+    border: 2px solid #ddd;
+    padding: 10px 10px 0px 10px;
+  }
+
+  .responses-title {
+    margin-bottom: 5px;
```
saharm1: remove

freetextresponse/private/view.less
```diff
@@ -111,5 +111,50 @@
...
+  }
+
+  .other-student-responses {
+    display: inline-block;
+    background: white;
+    width: 97%;
+    color: #414141;
+    text-align: left;
+    padding: 10px;
+    margin-bottom: 10px;
```
saharm1: margin-top: 10px

freetextresponse/private/view.less
```diff
@@ -111,5 +111,50 @@
     font-size: 0.9em;
   }
 
+  .responses-box {
+    display: none;
+    margin-top: 10px;
+    background: #f8f8f8;
+    border: 2px solid #ddd;
+    padding: 10px 10px 0px 10px;
```
saharm1: just 10px

freetextresponse/private/view.less
```diff
@@ -111,5 +111,50 @@
     font-size: 0.9em;
   }
 
+  .responses-box {
+    display: none;
+    margin-top: 10px;
+    background: #f8f8f8;
+    border: 2px solid #ddd;
```
saharm1: 1px

freetextresponse/private/view.less
```diff
@@ -111,5 +111,50 @@
...
+    padding: 10px 10px 0px 10px;
+  }
+
+  .responses-title {
+    margin-bottom: 5px;
+  }
+
+  .hide-button {
```
saharm1: box-shadow:none;

freetextresponse/private/view.less
```diff
@@ -111,5 +111,50 @@
...
+    float: right;
+    background: none;
+    border: none;
+    padding: 0;
+    color: #069;
+    text-decoration: underline;
+    cursor: pointer;
```
saharm1: remove

freetextresponse/private/view.less
```diff
@@ -111,5 +111,50 @@
...
+    font-size: 0.8em;
```
saharm1: remove

freetextresponse/private/view.js
```diff
@@ -74,6 +84,22 @@ function FreeTextResponseView(runtime, element) {
...
     });
 
+    function htmlBuilder(responses) {
+        /*
+        Convert list of responses to a html string to add to the page
+        */
+        var html = '';
+        //for each item in the list, add a div containing it to the html
+        responses.forEach(function(item) {
+            html += '<div class="other-student-responses">' + item.answer + '</div>';
```
saharm1: li not div

freetextresponse/private/view.less
```diff
@@ -111,5 +111,50 @@
...
+  }
+
+  .response-list {
+    list-style: none;
+    text-align: center;
```
saharm1: can probably remove

freetextresponse/templates/freetextresponse_view.html
```diff
@@ -23,4 +25,9 @@ <h2 class="problem-header">{{ display_name }}</h2>
         <div class="used-attempts-feedback" aria-live="polite">{{ used_attempts_feedback }}</div>
     </div>
     <div class="capa_alert user_alert">{{ user_alert }}</div>
+    <div class="responses-box" id="responses-box">
+        <button class="hide-button">Hide</button>
+        <p class="responses-title">Submissions by others</p>
+        <ul class="response-list">No responses to show at this time</ul>
```
saharm1: p tag

caesar2164 (approved):

freetextresponse/answer_pool.py
```diff
@@ -0,0 +1,58 @@
+"""
+Represents interaction with the answer pool, where student responses are stored
+"""
+import sys
+
+# Copied from https://github.com/ubc/ubcpi/blob/master/ubcpi/answer_pool.py
+# pool size in bytes, the whole pool will be stored in the block and retrieved
+# each time the page loads so we want to keep the pool to a reasonable small
+# size.
+POOL_SIZE = 4096
```
stvstnfrd: Have you inspected this w/ debug/print statements?
I wonder how large the pool actually gets...

Maybe add a `print(sys.getsizeof(pool))` after the `pool.append` call to monitor it, then:
- add 4+ responses to the pool, watch as pool size changes
- add 4+ responses of the maximum length (what's the largest response you can create? how many characters?), watch as pool size changes.

freetextresponse/templates/freetextresponse_view.html
```diff
@@ -23,4 +25,11 @@ <h2 class="problem-header">{{ display_name }}</h2>
         <div class="used-attempts-feedback" aria-live="polite">{{ used_attempts_feedback }}</div>
     </div>
     <div class="capa_alert user_alert">{{ user_alert }}</div>
+    <div class="responses-box" id="responses-box">
+        <button class="hide-button">Hide</button>
+        <p class="responses-title">Submissions by others</p>
+        <ul class="response-list">
+            <p class="no-response">No responses to show at this time</p>
```
stvstnfrd: Shouldn't this be an `<LI>`?

freetextresponse/templates/freetextresponse_view.html
```diff
@@ -23,4 +25,11 @@ <h2 class="problem-header">{{ display_name }}</h2>
         <div class="used-attempts-feedback" aria-live="polite">{{ used_attempts_feedback }}</div>
     </div>
     <div class="capa_alert user_alert">{{ user_alert }}</div>
+    <div class="responses-box" id="responses-box">
```
stvstnfrd: Do you need the ID? Can it be dropped in favor of the class?

freetextresponse/templates/freetextresponse_view.html
```diff
@@ -10,6 +10,8 @@ <h2 class="problem-header">{{ display_name }}</h2>
             <span class="status {{ visibility_class }}" aria-describedby="student_answer"></span>
         </div>
     </div>
+    <input class="messageCheckbox" type="checkbox" checked>Allow my response
```
stvstnfrd: TODO: Steven to write more about formatting here..

freetextresponse/public/view.less
```diff
@@ -111,5 +111,51 @@
     font-size: 0.9em;
   }
 
+  .responses-box {
```
stvstnfrd: TODO: Should we convert this directly to a CSS file and remove the `less` preprocessing?

freetextresponse/freetextresponse.py
```diff
@@ -269,6 +358,9 @@ def student_view(self, context={}):
                 'used_attempts_feedback': self._get_used_attempts_feedback(),
                 'visibility_class': self._get_indicator_visibility_class(),
                 'word_count_message': self._get_word_count_message(),
+                'display_other_responses':
+                    self.display_other_student_responses,
```
stvstnfrd: Eh, even if it's long, combine this with the line above;
it reads weird otherwise.

freetextresponse/freetextresponse.py
```diff
@@ -86,6 +152,28 @@ def workbench_scenarios():
...
+        default=False,
+        scope=Scope.settings,
+    )
+    system_answers = List(
+        default=[],
+        scope=Scope.user_state_summary,
+        help=_("System selected answers to give to students"),
+    )
+    other_responses_shown = List(
+        default=[],
```
stvstnfrd: Will this be problematic?
Will the same object be used as the default for all?
Where else are we doing this?

freetextresponse/freetextresponse.py
```diff
@@ -599,6 +699,22 @@ def save_reponse(self, data, suffix=''):
         }
         return result
 
+    def store_student_response(self, answer, record_response=False):
+        """
+        Store response from students to the backend
+        """
+        student_item = self.get_student_item_dict()
+        self.other_responses_shown = get_other_answers(
```
stvstnfrd: Let's copy what we're doing w/ `offer_answer` below and just pass in `student_id` directly.
That way our `get_other_answers` method doesn't need to know anything about how that underlying data structure works (I mean, it's just a `dict`, but still, knowing what key to use, etc.) since all it cares about is checking the user_id.

And since we're referencing that key twice, let's memoize it to a variable:

```python
student_id = student_item['student_id']
self.other_responses_shown = get_other_answers(
    self.system_answers,
    student_id
)
if record_response:
    offer_answer(
        self.system_answers,
        answer,
        student_id
    )
```

stvstnfrd: Side note: is `student_id` the same as `user_id`?
Are the different ideas?

freetextresponse/answer_pool.py
```diff
@@ -0,0 +1,58 @@
...
+    """
+    pool.append({
+        'student_id': student_id,
+        'answer': answer,
+    })
+    while sys.getsizeof(pool) > POOL_SIZE:
+        del pool[0]
+
+
+def get_other_answers(pool, get_student_item_dict):
```
stvstnfrd: Per comment above, this should become `student_id`, instead of `get_student_item_dict`.

freetextresponse/answer_pool.py
```diff
@@ -0,0 +1,58 @@
...
+    """
+    Returns at most the lastest preset number of answers from the pool.
+    Does not return answers the student had submitted.
+    """
+    return_list = []
+    student_id = get_student_item_dict['student_id']
```
stvstnfrd: Then we won't need this.

freetextresponse/answer_pool.py
```diff
@@ -0,0 +1,58 @@
...
+
+    for answer in reversed(pool):
```
stvstnfrd: TODO: Steven to comment more.

freetextresponse/answer_pool.py
```diff
@@ -0,0 +1,58 @@
...
+        'student_id': student_id,
+        'answer': answer,
+    })
+    while sys.getsizeof(pool) > POOL_SIZE:
+        del pool[0]
+
+
+def get_other_answers(pool, get_student_item_dict):
+    """
+    Returns at most the lastest preset number of answers from the pool.
```
stvstnfrd: Format docstrings like an commit message:

```
Provide a single line summary

It can be accompanied by additional paragraph(s) of details,
separated from the "subject" with an empty line.
Much like git and github provide tooling that expects this format,
docstring tools use this blank line to split subject/body.
```

freetextresponse/freetextresponse.py
```diff
@@ -18,11 +18,77 @@
...
+    Copied from
+    https://github.com/edx/edx-ora2/blob/master/openassessment/xblock/openassessmentblock.py
+    """
+    def get_course_id(self):
+        """
+        Returns serialized course id
+        """
+        return self._serialize_opaque_key(self.xmodule_runtime.course_id)
+
+    def get_student_item_dict(self, anonymous_user_id=None):
```
stvstnfrd: The more I look at it, it seems like we're not really using the `student_item_dict` for much more than the `student_id`, right? I don't see other uses here presently.

If that's the case, let's remove this method everywhere and replace it with `get_student_id` or what not..

freetextresponse/freetextresponse.py
```diff
@@ -18,11 +18,77 @@
...
+            student_id = unicode(student_id)
+        student_item_dict = {
+            'student_id': student_id,
+            'item_id': item_id,
+            'course_id': course_id,
+            'item_type': 'freetextresponse'
+        }
+        return student_item_dict
+
+    def get_user_role(self):
```
stvstnfrd: It looks like this is unused?

freetextresponse/freetextresponse.py
```diff
@@ -18,11 +18,77 @@
 from xblock.fragment import Fragment
 from xblock.validation import ValidationMessage
 from xblockutils.studio_editable import StudioEditableXBlockMixin
+from .answer_pool import offer_answer, get_other_answers
 from .mixins import EnforceDueDates
 
 
+class MissingDataFetcherMixin(object):
```
stvstnfrd: We should define this mixin alongside `EnforceDueDates` in `mixins.py`.
Then we can import it here.

freetextresponse/freetextresponse.py
```diff
@@ -18,11 +18,77 @@
...
+        Serialize key
+        """
+        if hasattr(key, 'to_deprecated_string'):
+            return key.to_deprecated_string()
+        return unicode(key)
+
+
 @XBlock.needs("i18n")
-class FreeTextResponse(EnforceDueDates, StudioEditableXBlockMixin, XBlock):
+class FreeTextResponse(EnforceDueDates, MissingDataFetcherMixin,
```
stvstnfrd: If we can't fit them all on a single line, break them up one per line:

```python
class FreeTextResponse(
        EnforceDueDates,
        MissingDataFetcherMixin,
        StudioEditableXBlockMixin,
        XBlock,
):
    pass
```

freetextresponse/freetextresponse.py
```diff
@@ -564,6 +656,12 @@ def submit(self, data, suffix=''):
             # even if word count is invalid.
             self.count_attempts += 1
             self._compute_score()
+            # check if valid
```
stvstnfrd: Are we checking if anything is valid _here_? or inside of `store_student_response`?

freetextresponse/freetextresponse.py
```diff
@@ -86,6 +152,28 @@ def workbench_scenarios():
...
+        scope=Scope.settings,
+    )
+    system_answers = List(
+        default=[],
+        scope=Scope.user_state_summary,
+        help=_("System selected answers to give to students"),
+    )
+    other_responses_shown = List(
+        default=[],
+        scope=Scope.user_state,
```
stvstnfrd: I have some questions about this.
Let's go over it tomorrow.

freetextresponse/private/view.js
```diff
@@ -47,7 +52,8 @@ function FreeTextResponseView(runtime, element) {
         $.ajax(url, {
             type: 'POST',
             data: JSON.stringify({
-                'student_answer': $element.find('.student_answer').val()
+                'student_answer': $element.find('.student_answer').val(),
+                'checked': $element.find('.messageCheckbox').prop('checked')
```
stvstnfrd: Can we call this variable something a bit more specific?
Something that expresses what it represents.
Maybe `can_record_response` or some such?

freetextresponse/private/view.js
```diff
@@ -57,12 +63,19 @@ function FreeTextResponseView(runtime, element) {
                 buttonSubmit.text(buttonSubmit[0].dataset.value);
                 userAlertMessage.text(response.user_alert);
                 buttonSave.addClass(response.nodisplay_class);
-                setClassForTextAreaParent(response.indicator_class); 
+                setClassForTextAreaParent(response.indicator_class);
+                if (!response.user_alert && response.display_other_responses) {
```
stvstnfrd: Why do we check the user alert?
I'm not familiar with what that is.

freetextresponse/private/view.js
```diff
@@ -74,6 +87,18 @@ function FreeTextResponseView(runtime, element) {
         return false;
     });
 
+    function htmlBuilder(responses) {
```
stvstnfrd: The name `htmlBuilder` gives the impression that this is a class or some object.
It's also not clear _what_ kind of html it builds.
Can we call it something like `get_student_responses_html` so that it's clear:
- what it does; it _gets_ something... html
- what kind of thing it gets it for, the student responses

freetextresponse/private/view.js
```diff
@@ -74,6 +87,18 @@ function FreeTextResponseView(runtime, element) {
...
+        /*
+        Convert list of responses to a html string to add to the page
+        */
+        var html = '';
+        //for each item in the list, add a div containing it to the html
```
stvstnfrd: It looks like this is no longer accurate?
Either way, I think we can remove because the code is pretty clear on what it's doing.

freetextresponse/public/view.js
```diff
@@ -4,6 +4,7 @@ function FreeTextResponseView(runtime, element) {
     var $ = window.jQuery;
     var $element = $(element);
     var $xblocksContainer = $('#seq_content');
+    var buttonHide = $element.find('.hide-button');
```
stvstnfrd, line 7: Let's talk about replacing the public file w/ the private file here too..

stvstnfrd (changes requested):
Okay, a few things this time.
I also have some other questions about how we're storing the answers; let's talk more about that tomorrow.

freetextresponse/answer_pool.py
saharm1: The pool gets nowhere near this size. I added 20 pretty long answers and the print statement only showed the size to be around 300. Additionally, the pool size changes suspiciously: sometimes I would add a response and the size would not go up. Can you explain why that would be the case?

freetextresponse/private/view.js
saharm1: We check the user alert because submit still goes through if there are issues like invalid word count. I can also do this in the python- will that be more logical?

freetextresponse/answer_pool.py
```diff
@@ -0,0 +1,59 @@
...
+                ...
+            ]
+        answer: the answer the student gave
+        student_id: student identifier
+    """
+    pool.append({
+        'student_id': student_id,
+        'answer': answer,
+    })
+    print(sys.getsizeof(pool))
```
saharm1: remove!!

freetextresponse/answer_pool.py
```diff
@@ -0,0 +1,59 @@
...
+                {
+                    ...
+                    ...
+                }
+                ...
+            ]
+        answer: the answer the student gave
+        student_id: student identifier
+    """
+    pool.append({
```
stvstnfrd: Before this, check if we have a response for this user. If so, remove it.

freetextresponse/answer_pool.py
```diff
@@ -0,0 +1,59 @@
...
+        'student_id': student_id,
+        'answer': answer,
+    })
+    print(sys.getsizeof(pool))
+    while sys.getsizeof(pool) > POOL_SIZE:
```
stvstnfrd: Remove this now.

freetextresponse/freetextresponse.py
stvstnfrd: Per f2f, this seems fine.

freetextresponse/templates/freetextresponse_view.html
saharm1: p tag

freetextresponse/templates/freetextresponse_view.html
```diff
@@ -23,4 +25,11 @@ <h2 class="problem-header">{{ display_name }}</h2>
         <div class="used-attempts-feedback" aria-live="polite">{{ used_attempts_feedback }}</div>
     </div>
     <div class="capa_alert user_alert">{{ user_alert }}</div>
+    <div class="responses-box">
+        <button class="hide-button">Hide</button>
+        <p class="responses-title">Submissions by others</p>
+        <ul class="response-list">
+            <p class="no-response">No responses to show at this time</p>
```
saharm1: li not p

stvstnfrd: Try to make this an `<LI>`. If needed, use the `no-response` class to remove LI bullet.

freetextresponse/answer_pool.py
```diff
@@ -0,0 +1,59 @@
...
+
+    Does not return answers the student had submitted.
+    """
+    return_list = []
+
+    for answer in reversed(pool):
+        if len(return_list) >= MAX_RESPONSES:
+            break
+        if not answer['student_id'] == student_id:
+            return_list.append(answer)
```
stvstnfrd:
```python
# use a list comprehension
def get_other_answers(pool, student_id):
    data = [
        answer
        for answer in reversed(pool):
        if student_id != answer['student_id']:
    ]
    data = data[:MAX_RESPONSES]
    return data
```

freetextresponse/answer_pool.py
```diff
@@ -0,0 +1,59 @@
...
+        answer: the answer the student gave
+        student_id: student identifier
+    """
+    pool.append({
+        'student_id': student_id,
+        'answer': answer,
+    })
+    print(sys.getsizeof(pool))
+    while sys.getsizeof(pool) > POOL_SIZE:
+        del pool[0]
```
stvstnfrd: Per 2f2, I think the new version of this should look something like:

```python
def offer_answer(pool, answer, student_id):
    # if self.score >= Credit.value.full:
    #    return pool
    data = [
        item
        for item in pool
        if student_id != item['student_id']
    ]
    data.append({
        'student_id': student_id,
        'answer': answer,
    })
    # Add space for one extra response
    # in case we have to filter out a user's own answer
    data = data[:MAX_RESPONSES+1]
    return data
```

Note: Instead of modifying the existing list (`pool`) in place,
let's return and then reassign a new collection.

saharm1:
@stvstnfrd I believe I have committed the changes we discussed. I also tried changing the scope of system answers to scope.settings, but it failed when I tried to edit it with this message: raise InvalidScopeError("{block}.{name} is read-only, cannot set".format(block=block, name=name)). I undid it for the time being.

freetextresponse/freetextresponse.py
```diff
@@ -18,11 +19,28 @@
 from xblock.fragment import Fragment
 from xblock.validation import ValidationMessage
 from xblockutils.studio_editable import StudioEditableXBlockMixin
-from .mixins import EnforceDueDates
+from .answer_pool import offer_answer, get_other_answers
+from .mixins import EnforceDueDates, MissingDataFetcherMixin
+
+
+MAX_RESPONSES = 3
```
saharm1: should be defined as field of xblock, not outside

saharm1: should this be defined as a field of the xblock?

freetextresponse/freetextresponse.py
```diff
@@ -18,11 +19,27 @@
 from xblock.fragment import Fragment
 from xblock.validation import ValidationMessage
 from xblockutils.studio_editable import StudioEditableXBlockMixin
-from .mixins import EnforceDueDates
+from .mixins import EnforceDueDates, MissingDataFetcherMixin
+
+
+MAX_RESPONSES = 3
```
saharm1, line 25: Should this be defined as a field of the XBlock?

stvstnfrd: This is fine for now. We can always move it later.
Once it's there, it's harder to get rid of, so let's not do it until we need it.

freetextresponse/freetextresponse.py
```diff
@@ -18,11 +19,27 @@
...
+    return data.decode('utf8')
 
 
 @XBlock.needs("i18n")
-class FreeTextResponse(EnforceDueDates, StudioEditableXBlockMixin, XBlock):
+class FreeTextResponse(
+    EnforceDueDates,
+    MissingDataFetcherMixin,
+    StudioEditableXBlockMixin,
+    XBlock
```
stvstnfrd: Add a trailing comma?

freetextresponse/freetextresponse.py
```diff
@@ -86,6 +103,20 @@ def workbench_scenarios():
...
+            'This will display other student responses to the '
+            'student after they submit their response.'
+        ),
+        default=False,
+        scope=Scope.settings,
+    )
+    system_answers = List(
+        default=[],
+        scope=Scope.user_state_summary,
+        help=_("System selected answers to give to students"),
```
stvstnfrd: Single quotes for consistency.

freetextresponse/freetextresponse.py
```diff
@@ -269,20 +291,26 @@ def student_view(self, context={}):
...
-            initialize_js_func='FreeTextResponseView',
-            additional_css=[
-                'public/view.less.min.css',
-            ],
-            additional_js=[
-                'public/view.js.min.js',
-            ],
         )
+        fragment.add_css(
+            _resource_string(
```
stvstnfrd: Was there a reason you departed from the `self.runtime.local_resource_url` calls in `build_fragment`?

Previously, the HTML page loaded CSS/JS via external files.
This changes that so that each page load reads the css/js file into memory and then includes the entire contents in each response (inlined into the html).

This is slower and requires more memory for the server, while increasing the data usage for clients/users. Especially for people on mobile or with otherwise limited internet access, we should let the browser cache these kinds of things, which it can do as long as we load from external files via URL, instead of inlining code.

freetextresponse/templates/freetextresponse_view.html
```diff
@@ -23,4 +27,11 @@ <h2 class="problem-header">{{ display_name }}</h2>
         <div class="used-attempts-feedback" aria-live="polite">{{ used_attempts_feedback }}</div>
     </div>
     <div class="capa_alert user_alert">{{ user_alert }}</div>
+    <div class="responses-box">
```
stvstnfrd: Shouldn't this be wrapped in an `if` block that checks if the feature is enabled for this problem?

freetextresponse/templates/freetextresponse_view.html
```diff
@@ -10,6 +10,10 @@ <h2 class="problem-header">{{ display_name }}</h2>
             <span class="status {{ visibility_class }}" aria-describedby="student_answer"></span>
         </div>
     </div>
+    <input class="messageCheckbox" type="checkbox" checked>
```
stvstnfrd: Shouldn't this be wrapped in an `if` block that checks if the feature is enabled for this problem?

freetextresponse/public/view.css
```diff
@@ -0,0 +1,155 @@
...
+.freetextresponse .user_input.unanswered .student_answer {
+  border:2px solid #e4e4e4;
+}
+
+.freetextresponse .user_input.correct .student_answer {
+  border:2px solid #1e9348;
+}
+
+.freetextresponse .user_input.incorrect .student_answer {
+  border:2px solid #b20610;
```
stvstnfrd: Here and throughout the CSS, follow style guidelines of including a space after the colon `:`.

freetextresponse/mixins.py
```diff
@@ -31,3 +31,35 @@ def is_past_due(self):
...
+        """
+        Get the student id.
+
+        See also: submissions.api for details.
+
+        Args:
+            anonymous_user_id(str): A unique anonymous_user_id for (user,
+            course) pair.
+        Returns:
+            (dict): The student item associated with this XBlock instance. This
```
stvstnfrd: This is no longer accurate, right?

freetextresponse/mixins.py
```diff
@@ -31,3 +31,35 @@ def is_past_due(self):
...
+    The mixin used for generating student item dict for submission API
+
+    Copied from
+    https://github.com/edx/edx-ora2/blob/master/openassessment/xblock/openassessmentblock.py
+    """
+    def get_student_id(self, anonymous_user_id=None):
+        """
+        Get the student id.
+
+        See also: submissions.api for details.
```
stvstnfrd: This is no longer accurate, right?

freetextresponse/mixins.py
```diff
@@ -31,3 +31,35 @@ def is_past_due(self):
...
+class MissingDataFetcherMixin(object):
+    """
+    The mixin used for generating student item dict for submission API
+
+    Copied from
+    https://github.com/edx/edx-ora2/blob/master/openassessment/xblock/openassessmentblock.py
+    """
+    def get_student_id(self, anonymous_user_id=None):
+        """
+        Get the student id.
```
stvstnfrd: Honestly, we could delete the rest of this docstring and just leave this line; it's simple enough.

freetextresponse/mixins.py
```diff
@@ -31,3 +31,35 @@ def is_past_due(self):
...
             return now > due
         return False
+
+
+class MissingDataFetcherMixin(object):
+    """
+    The mixin used for generating student item dict for submission API
+
+    Copied from
+    https://github.com/edx/edx-ora2/blob/master/openassessment/xblock/openassessmentblock.py
```
stvstnfrd: This is no longer accurate, right?
We've departed enough that it's no longer the same code.

freetextresponse/mixins.py
```diff
@@ -31,3 +31,35 @@ def is_past_due(self):
                 due = due + graceperiod
             return now > due
         return False
+
+
+class MissingDataFetcherMixin(object):
+    """
+    The mixin used for generating student item dict for submission API
```
stvstnfrd: This is no longer accurate, right?

freetextresponse/mixins.py
```diff
@@ -31,3 +31,35 @@ def is_past_due(self):
...
+
+    Copied from
+    https://github.com/edx/edx-ora2/blob/master/openassessment/xblock/openassessmentblock.py
+    """
+    def get_student_id(self, anonymous_user_id=None):
```
stvstnfrd: We're not using the `anonymous_user_id` kwarg, so I think we can just delete it..

freetextresponse/mixins.py
```diff
@@ -31,3 +31,35 @@ def is_past_due(self):
...
+        """
+        Get the student id.
+
+        See also: submissions.api for details.
+
+        Args:
+            anonymous_user_id(str): A unique anonymous_user_id for (user,
```
stvstnfrd: See note above about maybe dropping this.

freetextresponse/mixins.py
```diff
@@ -31,3 +31,35 @@ def is_past_due(self):
...
+            course) pair.
+        Returns:
+            (dict): The student item associated with this XBlock instance. This
+            includes the student id, item id, and course id.
+        """
+        if hasattr(self, "xmodule_runtime"):
```
stvstnfrd: Single quotes for consistency.

freetextresponse/freetextresponse.py
```diff
@@ -18,11 +19,27 @@
...
 from xblock.validation import ValidationMessage
 from xblockutils.studio_editable import StudioEditableXBlockMixin
-from .mixins import EnforceDueDates
+from .mixins import EnforceDueDates, MissingDataFetcherMixin
+
+
+MAX_RESPONSES = 3
+
+
+def _resource_string(path):
```
stvstnfrd: See note below about dropping this.
If so, let's remove this.

freetextresponse/freetextresponse.py
```diff
@@ -86,6 +103,20 @@ def workbench_scenarios():
...
+    display_other_student_responses = Boolean(
+        display_name=_('Display Other Student Responses'),
+        help=_(
+            'This will display other student responses to the '
+            'student after they submit their response.'
+        ),
+        default=False,
+        scope=Scope.settings,
+    )
+    system_answers = List(
```
stvstnfrd: Is there a more descriptive name we can use?
- `provided_answers`
- `sampled_answers`
- `displayable_answers`

Thoughts?

freetextresponse/freetextresponse.py
```diff
@@ -269,20 +291,26 @@ def student_view(self, context={}):
...
                 'word_count_message': self._get_word_count_message(),
+                'display_other_responses': self.display_other_student_responses,
+                'other_responses': [],
             }
         )
         template = get_template('freetextresponse_view.html')
         fragment = self.build_fragment(
             template,
             context,
-            initialize_js_func='FreeTextResponseView',
```
stvstnfrd: See note below about restoring this..

freetextresponse/freetextresponse.py
```diff
@@ -599,6 +635,76 @@ def save_reponse(self, data, suffix=''):
         }
         return result
 
+    def store_student_response(self, answer, record_response=True):
+        """
+        Store response from students to the backend
+        """
+        student_id = self.get_student_id()
+        if record_response:
+            self.offer_answer(
```
stvstnfrd: You know, now that we've moved things around, I don't think we really need this to be split out into its own method. If the callee already knows `record_response`, it should just decide whether or not to call this method. At that point, there's not much left in this method, so let's just move `offer_answer` up inline into `store_student_response`.

freetextresponse/freetextresponse.py
```diff
@@ -564,6 +592,12 @@ def submit(self, data, suffix=''):
             # even if word count is invalid.
             self.count_attempts += 1
             self._compute_score()
+            # check if instructor wants to display other responses
+            if self.display_other_student_responses:
+                self.store_student_response(
+                    self.student_answer,
+                    record_response=data['can_record_response']
```
stvstnfrd: I guess if we already know right here if we're allowed to record their response, maybe we should just check that too in the `if` clause above?
Seems unnecessary to call the method while also telling it whether or not it's allowed to do it :)

```python
if self.display_other_student_responses and data['can_record_response']:
    self.store_student_response()
```

stvstnfrd: We also don't need to pass in an answer since that's available via `self` already.

freetextresponse/freetextresponse.py
```diff
@@ -564,6 +592,12 @@ def submit(self, data, suffix=''):
             # even if word count is invalid.
             self.count_attempts += 1
             self._compute_score()
+            # check if instructor wants to display other responses
```
stvstnfrd: Since the `if` check has evolved, we may want to just drop this comment instead of worrying about keeping it in sync. The `if` conditions are pretty clear already.

freetextresponse/freetextresponse.py
stvstnfrd:
```python
def store_student_response(self):
    # if the answer is wrong, do not display it
    if self.score != Credit.full.value:
        return
    student_id = self.get_student_id()
    answer = self.student_answer
    for index, response in enumerate(self.system_answers):
        if response['student_id'] == student_id:
            del self.system_answers[index]
            break
    # etc...
```

freetextresponse/freetextresponse.py
```diff
@@ -599,6 +635,76 @@ def save_reponse(self, data, suffix=''):
...
+
+        self.system_answers.append({
+            'student_id': student_id,
+            'answer': answer,
+        })
+
+        # Want to store extra response so student can still see
+        # MAX_RESPONSES answers if their answer is in the pool.
+        if len(self.system_answers) > MAX_RESPONSES + 1:
+            del self.system_answers[0]
```
stvstnfrd: While hopefully unlikely, if two extra answers were ever added to the answer list, we wouldn't ever clean it up, since this is written to trim at most one item, never more.
Instead, if we use a list slice, we can grab the exact subset we want in a single line.

```python
self.system_answers = self.system_answers[-(MAX_RESPONSES+1):]
```

By using a negative value, the slice counts backward from the end of the string in determining where to begin. And since there's nothing after the colon, it grabs until the end of the string.
This behaves no matter the size of the list, even empty.
In our case, it translates to "move back up to 4 items back starting from the end of the list, then grab everything between there and the end."
We don't even need an `if` check!

freetextresponse/freetextresponse.py
```diff
@@ -599,6 +635,76 @@ def save_reponse(self, data, suffix=''):
...
+        Does not return answers the student had submitted.
+        """
+        return_list = [
+            response
+            for response in self.system_answers
+            if response['student_id'] != student_id
+        ]
+
+        if len(return_list) > MAX_RESPONSES:
+            del return_list[0]
```
stvstnfrd: See note above about making this a list slice to prune off any extras.

```python
return_list = return_list[-(MAX_RESPONSES):]
```

stvstnfrd (changes requested):
A few style things, some final tidying up.. almost there!

stvstnfrd (approved):
Looks great! Maybe give me one me quick demo after standup, but I think we're ready to go!

PR #33, opened 2018-09-25
Show Peer Responses on page load if q answered

caesar2164:
- on page load make ajax call with student answer
- if student answer is valid display peer responses
- also defines a function to check for response validity and displaying peer responses that gets used on both page load and submit click

@stvstnfrd this was remarkably easy...

### DIFF
--- a/freetextresponse/freetextresponse.py
+++ b/freetextresponse/freetextresponse.py
@@ -279,7 +279,6 @@ class FreeTextResponse(
             (Fragment): The HTML Fragment for this XBlock, which determines the
             general frame of the FreeTextResponse Question.
         """
-
         self.runtime.service(self, 'i18n')
         context.update(
             {
@@ -294,7 +293,7 @@ class FreeTextResponse(
                 'visibility_class': self._get_indicator_visibility_class(),
                 'word_count_message': self._get_word_count_message(),
                 'display_other_responses': self.display_other_student_responses,
-                'other_responses': [],
+                'other_responses': self.get_other_answers(),
             }
         )
         template = get_template('freetextresponse_view.html')
@@ -602,7 +601,7 @@ class FreeTextResponse(
             'user_alert': self._get_user_alert(
                 ignore_attempts=True,
             ),
-            'other_responses': self.get_other_answers(self.get_student_id()),
+            'other_responses': self.get_other_answers(),
             'display_other_responses': self.display_other_student_responses,
             'visibility_class': self._get_indicator_visibility_class(),
         }
@@ -654,12 +653,19 @@ class FreeTextResponse(
         # MAX_RESPONSES answers if their answer is in the pool.
         self.displayable_answers = self.displayable_answers[-(MAX_RESPONSES+1):]
 
-    def get_other_answers(self, student_id):
+    def get_other_answers(self):
         """
         Returns at most MAX_RESPONSES answers from the pool.
 
         Does not return answers the student had submitted.
         """
+        student_id = self.get_student_id()
+        display_other_responses = self.display_other_student_responses
+        shouldnt_show_other_responses = not display_other_responses
+        student_answer_incorrect = self._determine_credit() == Credit.zero
+        if student_answer_incorrect or shouldnt_show_other_responses:
+            return []
+
         return_list = [
             response
             for response in self.displayable_answers
--- a/freetextresponse/freetextresponse.py
+++ b/freetextresponse/freetextresponse.py
@@ -2,7 +2,6 @@
 This is the core logic for the Free-text Response XBlock
 """
 from enum import Enum
-import pkg_resources
 from django.db import IntegrityError
 from django.template.context import Context
 from django.template.loader import get_template
@@ -27,10 +26,10 @@ MAX_RESPONSES = 3
 
 @XBlock.needs("i18n")
 class FreeTextResponse(
-    EnforceDueDates,
-    MissingDataFetcherMixin,
-    StudioEditableXBlockMixin,
-    XBlock,
+        EnforceDueDates,
+        MissingDataFetcherMixin,
+        StudioEditableXBlockMixin,
+        XBlock,
 ):
     #  pylint: disable=too-many-ancestors, too-many-instance-attributes
     """
@@ -279,6 +278,9 @@ class FreeTextResponse(
             (Fragment): The HTML Fragment for this XBlock, which determines the
             general frame of the FreeTextResponse Question.
         """
+
+        display_other_responses = self.display_other_student_responses
+
         self.runtime.service(self, 'i18n')
         context.update(
             {
@@ -292,7 +294,7 @@ class FreeTextResponse(
                 'used_attempts_feedback': self._get_used_attempts_feedback(),
                 'visibility_class': self._get_indicator_visibility_class(),
                 'word_count_message': self._get_word_count_message(),
-                'display_other_responses': self.display_other_student_responses,
+                'display_other_responses': display_other_responses,
                 'other_responses': self.get_other_answers(),
             }
         )
@@ -589,7 +591,8 @@ class FreeTextResponse(
             # even if word count is invalid.
             self.count_attempts += 1
             self._compute_score()
-            if self.display_other_student_responses and data.get('can_record_response'):
+            display_other_responses = self.display_other_student_responses
+            if display_other_responses and data.get('can_record_response'):
                 self.store_student_response()
         result = {
             'status': 'success',
@@ -651,7 +654,8 @@ class FreeTextResponse(
 
         # Want to store extra response so student can still see
         # MAX_RESPONSES answers if their answer is in the pool.
-        self.displayable_answers = self.displayable_answers[-(MAX_RESPONSES+1):]
+        response_index = -(MAX_RESPONSES+1)
+        self.displayable_answers = self.displayable_answers[response_index:]
 
     def get_other_answers(self):
         """
@@ -665,7 +669,6 @@ class FreeTextResponse(
         student_answer_incorrect = self._determine_credit() == Credit.zero
         if student_answer_incorrect or shouldnt_show_other_responses:
             return []
-
         return_list = [
             response
             for response in self.displayable_answers
--- a/freetextresponse/mixins.py
+++ b/freetextresponse/mixins.py
@@ -1,10 +1,11 @@
 """
 Mixins for the Free Text Response XBlock
 """
+# pylint: disable=too-few-public-methods
 import datetime
 
 
-class EnforceDueDates(object):  # pylint: disable=too-few-public-methods
+class EnforceDueDates(object):
     """
     xBlock Mixin to allow xblocks to check the due date
     (taking the graceperiod into account) of the
--- a/freetextresponse/public/view.js
+++ b/freetextresponse/public/view.js
@@ -68,13 +68,7 @@ function FreeTextResponseView(runtime, element) {
                 userAlertMessage.text(response.user_alert);
                 buttonSave.addClass(response.nodisplay_class);
                 setClassForTextAreaParent(response.indicator_class);
-                if (!response.user_alert && response.display_other_responses) {
-                    var responseHTML = get_student_responses_html(response.other_responses);
-                    if (responseHTML) {
-                        responseList.html(responseHTML);
-                    }
-                    $element.find('.responses-box').show();
-                }
+                displayResponsesIfAnswered(response);
 
                 $xblocksContainer.data(cachedAnswerId, $element.find('.student_answer').val());
                 $xblocksContainer.data(problemProgressId, response.problem_progress);
@@ -91,15 +85,27 @@ function FreeTextResponseView(runtime, element) {
         return false;
     });
 
-    function get_student_responses_html(responses) {
+    function getStudentResponsesHtml(responses) {
         /*
         Convert list of responses to a html string to add to the page
         */
         var html = '';
+        var noResponsesText = responseList.data('noresponse');
         responses.forEach(function(item) {
             html += '<li class="other-student-responses">' + item.answer + '</li>';
         });
+        html = html || '<li class="no-response">' + noResponsesText + '</li>';
         return html;
+    }
+
+    function displayResponsesIfAnswered(response) {
+        if (!response.display_other_responses) {
+            $element.find('.responses-box').addClass('hidden');
+            return;
+        }
+        var responseHTML = getStudentResponsesHtml(response.other_responses);
+        responseList.html(responseHTML);
+        $element.find('.responses-box').removeClass('hidden');
     }
 
     buttonSave.on('click', function () {
--- a/freetextresponse/templates/freetextresponse_view.html
+++ b/freetextresponse/templates/freetextresponse_view.html
@@ -30,7 +30,7 @@
     </div>
     <div class="capa_alert user_alert">{{ user_alert }}</div>
     {% if display_other_responses %}
-        <div class="responses-box">
+        <div class="responses-box {% if not student_answer %}hidden{% endif %}">
             <button class="hide-button">
                 <span class="hide">Hide</span>
                 <span class="show">Show</span>
@@ -38,7 +38,11 @@
             </button>
             <p class="responses-title">Submissions by others</p>
             <ul class="response-list">
-                <li class="no-response">No responses to show at this time</li>
+                {% for response in other_responses %}
+                    <li class="other-student-responses">{{ response.answer }}</li>
+                {% empty %}
+                    <li class="no-response">No responses to show at this time</li>
+                {% endfor %}
             </ul>
         </div>
     {% endif %}
--- a/freetextresponse/templates/freetextresponse_view.html
+++ b/freetextresponse/templates/freetextresponse_view.html
@@ -12,8 +12,7 @@
     </div>
     {% if display_other_responses %}
         <input class="messageCheckbox" type="checkbox" checked>
-            <span>Allow my response to possibly be visible by other learners after
-            submitting their response</span>
+            <span>{% trans "Allow my response to possibly be visible by other learners after submitting their response" %}</span>
         <br>
     {% endif %}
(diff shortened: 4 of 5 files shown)


freetextresponse/public/view.js
```diff
@@ -139,4 +144,24 @@ function FreeTextResponseView(runtime, element) {
...
     });
+
+    $xblocksContainer.ready(function() {
+        $.ajax(url, {
+            type: 'POST',
+            data: JSON.stringify({
+                'student_answer': $element.find('.student_answer').val()
+            }),
+            success: function xblockLoadOnSuccess(response) {
+                display_responses_if_answered(response);
```
stvstnfrd: Instead of having two separate AJAX calls, can we refactor the two calls into a single helper?

Also, why are we waiting to make an AJAX request _after_ page load;
this means you're going to get a jump/flash of content when the ajax call finally goes through.

Since we already know at initial page load whether or not we should already display the results,
we should just include them directly in the original page load/response.

stvstnfrd (changes requested):

stvstnfrd:
@caesar2164 ping.

freetextresponse/public/view.js
caesar2164: I'm not sure if refactoring those two would net us any benefit right now. There are a bunch of differences both in data being sent, as well as in success reactions. (I already factored out the overlapping stuff into a helper)

Unfortunately we CAN'T do the AJAX request any earlier, XBlocks get loaded AFTER page load. Since I need to query the XBlock to get its answer, as well as peer responses, I HAVE to wait for the XBlock to be loaded like I do...

stvstnfrd: > Unfortunately we CAN'T do the AJAX request any earlier

I think you may have missed my point:

> we should just include them directly in the original page load/response.

I'm suggesting that you _not_ use AJAX _at_ _all_ for initializing the XBlock.

stvstnfrd (changes requested):

freetextresponse/public/view.js
caesar2164: The problem is that `response.other_responses` is only available in an AJAX response... so we CAN'T know if/what we should be displaying without an AJAX request.

caesar2164: ok, I guess we could, preload the HTML with answers, but I'm actually not sure if we should?

caesar2164:
nvmd, brain fart re AJAX vs page load...

freetextresponse/public/view.js
```diff
@@ -98,6 +93,16 @@ function FreeTextResponseView(runtime, element) {
         return html;
     }
 
+    function display_responses_if_answered(response) {
```
caesar2164: Should I remove this helper function?

caesar2164:
ok @stvstnfrd ready for round 2

freetextresponse/freetextresponse.py
```diff
@@ -279,6 +279,17 @@ def student_view(self, context={}):
...
             general frame of the FreeTextResponse Question.
         """
+        student_id = self.get_student_id()
+        display_other_responses = self.display_other_student_responses
+        student_answer = self.student_answer
+        if student_answer and display_other_responses:
+            other_responses = self.get_other_answers(student_id)
+        else:
+            other_responses = []
+        if other_responses:
```
stvstnfrd: Since this `if` check is purely presentational, shouldn't we move it to the template?

Instead of the backend telling the frontend whether or not to display an element,
we should instead delegate that decision to the template.

So then the template can rationalize,
"if I don't have other responses, I shouldn't try to display them".

freetextresponse/freetextresponse.py
```diff
@@ -279,6 +279,17 @@ def student_view(self, context={}):
             (Fragment): The HTML Fragment for this XBlock, which determines the
             general frame of the FreeTextResponse Question.
         """
+        student_id = self.get_student_id()
+        display_other_responses = self.display_other_student_responses
+        student_answer = self.student_answer
+        if student_answer and display_other_responses:
```
stvstnfrd: Kind of a nit, but I find it to be a little safer to explicitly declare a default value, instead of relying on the `else` clause. It also reads a bit more cleanly like:

```python
other_responses = []
if student_answer and display_other_responses:
    other_responses = self.get_other_answers(student_id)
```

freetextresponse/public/view.js
```diff
@@ -98,6 +93,16 @@ function FreeTextResponseView(runtime, element) {
...
+        if (!response.user_alert && response.display_other_responses) {
+            var responseHTML = get_student_responses_html(response.other_responses);
+            if (responseHTML) {
+                responseList.html(responseHTML);
+            }
+            $element.find('.responses-box').removeClass('hidden');
```
stvstnfrd: Should this instead be inside the `if` block?
That way it only tries to remove the class if it actually pulled back HTML?

Or actually, what happens if the AJAX lookup here (from `get_student_responses_html`) doesn't find any items. Should `get_student_responses_html` instead return the "No messages to display" instead of an empty string? In that case, we can remove the `if` check completely and just update the inner html regardless.

freetextresponse/public/view.js
stvstnfrd: I wouldn't.

Honestly, I tend to find that anything more than 5 (ish) lines can generally be pulled out into a dedicated function.

Here, this is a great level of abstraction; it's a helper function that makes it abundantly clear what's going on.
Otherwise, you end up needing to overly comment each stanza, something like:

Note: lines with `###` are placeholders for code, while single `#` are actually comments that would be added to the code

```python
def main():
    # initialize a bunch of stuff
    pass  ### ... lots of lines
    
    # fetch data
    pass  ### make ajax calls and whatever

    # parse data
    pass  ### several lines looping over data and processing it

    # display data
    pass  ### all the stuff in `display_responses_if_answered`
```

Now compare that to this version w/ helpers:

```python
def main():
    self.initialize_state()
    data = fetch_data()
    data = parse_data(data)
    display_responses_if_answered(data)
```

Well-named helper functions give us automatic documentation :)

freetextresponse/templates/freetextresponse_view.html
```diff
@@ -30,11 +30,15 @@ <h2 class="problem-header">{{ display_name }}</h2>
     </div>
     <div class="capa_alert user_alert">{{ user_alert }}</div>
     {% if display_other_responses %}
-        <div class="responses-box">
+        <div class="responses-box {{ hide_responses_class }}">
```
stvstnfrd: See note above about moving display logic into here.

stvstnfrd (changes requested):

freetextresponse/public/view.js
caesar2164: By default `responseList` is loaded with `<li class="no-response">No responses to show at this time</li>` if there isn't a student response. So if the AJAX call comes back empty it'll just show that default value...

caesar2164:
@stvstnfrd I think I've addressed everything. (I also rebased to resolve a conflict from the previous merged bugfix)

freetextresponse/public/view.js
stvstnfrd: I think that relies too much on assumptions of the underlying implementation.

If we don't have any `responseHTML` and we aren't updating the content of `responseList`,
then we shouldn't be modifying the responses box.

It's possible, however unexpected, that you could start with responses, and later get back nothing.
- a response could be deleted
- an error could occur that causes the lookup to fail and instead returns an empty list

We can't make this assumption.

An easy way to do this is to just bail immediately if there isn't anything to update:

```javascript
if (!responseHtml) {
    return;
}
responseList.html(responseHtml);
$element.find('.responses-box').removeClass('hidden');
```

But still, as pointed out, this doesn't address what happens if we _do_ have responses at one point, but then later we don't.
The easy way to handle this is to just have our `get_student_responses_html` helper return either the response html or the "no-response" html.

stvstnfrd: We can discuss more f2f, if needed.

freetextresponse/templates/freetextresponse_view.html
```diff
@@ -30,16 +30,24 @@ <h2 class="problem-header">{{ display_name }}</h2>
...
-                <span class="show">Show</span>
-                <span class="sr">peer responses</span>
-            </button>
-            <p class="responses-title">Submissions by others</p>
-            <ul class="response-list">
-                <li class="no-response">No responses to show at this time</li>
-            </ul>
-        </div>
+        {% if student_answer %}
+            <div class="responses-box">
```
stvstnfrd: I think this got over-engineered.
What you had here (class name in a variable) previously was fine.
In fact, it's what we want, to specify visibility with a variable.

I was just pointing out that it should be the template itself which assigns the variable,
not the backend.

freetextresponse/public/view.js
```diff
@@ -102,6 +97,16 @@ function FreeTextResponseView(runtime, element) {
         return html;
     }
 
+    function display_responses_if_answered(response) {
+        if (!response.user_alert && response.display_other_responses) {
+            var responseHTML = get_student_responses_html(response.other_responses);
```
stvstnfrd: And while we're in here, let's explicitly hoist the variable to the top of the function and update our naming to match JS convention (camelCase instead of snake_case).

`responseHtml` and `getStudentResponsesHtml`.

freetextresponse/public/view.js
stvstnfrd: Also, see note about JS using camelCase instead of snake_case.

freetextresponse/templates/freetextresponse_view.html
```diff
@@ -30,16 +30,24 @@ <h2 class="problem-header">{{ display_name }}</h2>
...
+                    <span class="hide">Hide</span>
+                    <span class="show">Show</span>
+                    <span class="sr">peer responses</span>
+                </button>
+                <p class="responses-title">Submissions by others</p>
+                <ul class="response-list">
+                    {% for response in other_responses %}
+                        <li class="other-student-responses">{{ response.answer }}</li>
+                    {% empty %}
+                        <li class="no-response">No responses to show at this time</li>
```
stvstnfrd: Does i18n happen automatically or does it need specified?

stvstnfrd (changes requested):

caesar2164:
@stvstnfrd ready for yet another round of review. (Diff down to +19!)

freetextresponse/freetextresponse.py
```diff
@@ -660,6 +660,12 @@ def get_other_answers(self, student_id):
 
         Does not return answers the student had submitted.
         """
+        display_other_responses = self.display_other_student_responses
+        shouldnt_show_other_responses = not display_other_responses
+        student_hasnt_answered_properly = self._determine_credit() == Credit.zero
+        if student_hasnt_answered_properly or shouldnt_show_other_responses:
+            return []
+
```
stvstnfrd: Nit: I'd remove the empty line..

freetextresponse/templates/freetextresponse_view.html
```diff
@@ -30,15 +30,19 @@ <h2 class="problem-header">{{ display_name }}</h2>
     </div>
     <div class="capa_alert user_alert">{{ user_alert }}</div>
     {% if display_other_responses %}
-        <div class="responses-box">
+        <div class="responses-box {% if not student_answer %}hidden{% endif %}">
```
stvstnfrd, line 32: Nice!

freetextresponse/freetextresponse.py
```diff
@@ -294,7 +294,7 @@ def student_view(self, context={}):
                 'visibility_class': self._get_indicator_visibility_class(),
                 'word_count_message': self._get_word_count_message(),
                 'display_other_responses': self.display_other_student_responses,
-                'other_responses': [],
+                'other_responses': self.get_other_answers(self.get_student_id()),
```
stvstnfrd: Since we're never re-using a memoized version of this, just move the `get_student_id` lookup into the method itself.

freetextresponse/public/view.js
```diff
@@ -102,6 +96,18 @@ function FreeTextResponseView(runtime, element) {
         return html;
     }
 
+    function display_responses_if_answered(response) {
+        if (!response.display_other_responses) {
+            $element.find('.responses-box').addClass('hidden');
+            return;
+        }
+        var responseHTML = get_student_responses_html(response.other_responses);
+        if (responseHTML) {
```
stvstnfrd: I have a question about this, I'll just discuss f2f..

stvstnfrd (changes requested):

freetextresponse/public/view.js
caesar2164: Resolved

stvstnfrd:
@caesar2164 What's the status of this?

caesar2164:
waiting for your review @stvstnfrd

freetextresponse/public/view.js
```diff
@@ -99,9 +93,20 @@ function FreeTextResponseView(runtime, element) {
         responses.forEach(function(item) {
             html += '<li class="other-student-responses">' + item.answer + '</li>';
         });
+        html = html || '<li class="no-response">No responses to show at this time</li>';
```
stvstnfrd: What about i18n?

freetextresponse/public/view.js
```diff
@@ -99,9 +93,20 @@ function FreeTextResponseView(runtime, element) {
...
         return html;
     }
 
+    function display_responses_if_answered(response) {
```
stvstnfrd: Javascript naming convention is camelCase not snake_case.

freetextresponse/public/view.js
```diff
@@ -99,9 +93,20 @@ function FreeTextResponseView(runtime, element) {
...
+        if (!response.display_other_responses) {
+            $element.find('.responses-box').addClass('hidden');
+            return;
+        }
+        var responseHTML = get_student_responses_html(response.other_responses);
```
stvstnfrd: Update style convention for JS: `responseHtml`...

stvstnfrd (changes requested):
Some small things, mostly style, otherwise looks good.

Can you do a final demo for Jen, Monica, and Kailey?

freetextresponse/mixins.py
```diff
@@ -33,7 +33,7 @@ def is_past_due(self):
         return False
 
 
-class MissingDataFetcherMixin(object):
+class MissingDataFetcherMixin(object):  # pylint: disable=too-few-public-methods
```
stvstnfrd: You should be able to also disable the long line warning here.

stvstnfrd:
@caesar2164 I resolved most of the convos, just a few notes left about i18n and pep8.

freetextresponse/public/view.js
caesar2164: unfortunately that doesn't work in xBlocks yet: https://openedx.atlassian.net/browse/PLAT-1141

stvstnfrd: @caesar2164 Then just stuff it into the HTML template as a hidden element and grab it to use as a template here.

freetextresponse/mixins.py
```diff
@@ -33,7 +33,7 @@ def is_past_due(self):
         return False
 
 
-class MissingDataFetcherMixin(object):
+class MissingDataFetcher(object):  # pylint: disable=too-few-public-methods
```
stvstnfrd: Eh, I don't know how I feel about renaming a sensible variable just because a pylint comment makes a line too long..

freetextresponse/templates/freetextresponse_view.html
```diff
@@ -30,15 +30,19 @@ <h2 class="problem-header">{{ display_name }}</h2>
...
                 <span class="show">Show</span>
                 <span class="sr">peer responses</span>
             </button>
             <p class="responses-title">Submissions by others</p>
             <ul class="response-list">
-                <li class="no-response">No responses to show at this time</li>
+                {% for response in other_responses %}
+                    <li class="other-student-responses">{{ response.answer }}</li>
+                {% empty %}
+                    <li class="no-response">No responses to show at this time</li>
```
stvstnfrd: Maybe always write this item out to the template, `display: none;` by default, like before,
then display it when when `len(other_responses) == 0`.

stvstnfrd: Sorry, you'd want to write it out so that you can grab it elsewhere in your javascript..

freetextresponse/freetextresponse.py
```diff
@@ -652,14 +654,21 @@ def store_student_response(self):
 
         # Want to store extra response so student can still see
         # MAX_RESPONSES answers if their answer is in the pool.
-        self.displayable_answers = self.displayable_answers[-(MAX_RESPONSES+1):]
+        response_key = -(MAX_RESPONSES+1)
```
stvstnfrd: This is a number, right?
Should it be `response_index`?

freetextresponse/freetextresponse.py
```diff
@@ -652,14 +654,21 @@ def store_student_response(self):
...
+    def get_other_answers(self):
         """
         Returns at most MAX_RESPONSES answers from the pool.
 
         Does not return answers the student had submitted.
         """
+        student_id = self.get_student_id()
+        display_other_responses = self.display_other_student_responses
+        shouldnt_show_other_responses = not display_other_responses
+        student_answer_not_proper = self._determine_credit() == Credit.zero
```
stvstnfrd: `student_answer_incorrect`?

"Proper" seems ambiguous...

freetextresponse/mixins.py
stvstnfrd: What if you move the disable to above the `class` declaration instead of beside it?

It's a little more broad, but really, it could apply to any/all classes in this file; that's fine.

stvstnfrd: And better yet, I think you can re-enable the flag again right after this class:

`# pylint: enable=too-foo-public-methods`

stvstnfrd (approved):

stvstnfrd (approved):

PR #35, opened 2018-11-18
Fix the FTR XBlock for Hawthorn

OmarIthawi:
The following error occurs when using this XBlock on Hawthorn.

### TODO

 - [x] Test also in Ficus (I don't have Ginkgo :man_shrugging:)

### Questions

 - [ ] How do I get the student responses?

### Before
<kbd>haw</kbd>

### After

This PR fixes it:

<kbd>haw-fixed</kbd>

### DIFF
--- a/freetextresponse/freetextresponse.py
+++ b/freetextresponse/freetextresponse.py
@@ -4,7 +4,6 @@ This is the core logic for the Free-text Response XBlock
 from enum import Enum
 from django.db import IntegrityError
 from django.template.context import Context
-from django.template.loader import get_template
 from django.utils.translation import ungettext
 from django.utils.translation import ugettext_lazy as _
 from django.utils.translation import ugettext
@@ -17,6 +16,7 @@ from xblock.fields import Scope
 from xblock.fields import String
 from xblock.fragment import Fragment
 from xblock.validation import ValidationMessage
+from xblockutils.resources import ResourceLoader
 from xblockutils.studio_editable import StudioEditableXBlockMixin
 from .mixins import EnforceDueDates, MissingDataFetcherMixin
 
@@ -35,6 +35,9 @@ class FreeTextResponse(
     """
     Enables instructors to create questions with free-text responses.
     """
+
+    loader = ResourceLoader(__name__)
+
     @staticmethod
     def workbench_scenarios():
         """
@@ -240,8 +243,7 @@ class FreeTextResponse(
 
     def build_fragment(
             self,
-            template,
-            context_dict,
+            rendered_template,
             initialize_js_func,
             additional_css=[],
             additional_js=[],
@@ -250,8 +252,7 @@ class FreeTextResponse(
         """
         Creates a fragment for display.
         """
-        context = Context(context_dict)
-        fragment = Fragment(template.render(context))
+        fragment = Fragment(rendered_template)
         for item in additional_css:
             url = self.runtime.local_resource_url(self, item)
             fragment.add_css_url(url)
@@ -278,9 +279,7 @@ class FreeTextResponse(
             (Fragment): The HTML Fragment for this XBlock, which determines the
             general frame of the FreeTextResponse Question.
         """
-
         display_other_responses = self.display_other_student_responses
-
         self.runtime.service(self, 'i18n')
         context.update(
             {
@@ -298,10 +297,13 @@ class FreeTextResponse(
                 'other_responses': self.get_other_answers(),
             }
         )
-        template = get_template('freetextresponse_view.html')
+        template = self.loader.render_django_template(
+            'templates/freetextresponse_view.html',
+            context=Context(context),
+            i18n_service=self.runtime.service(self, 'i18n'),
+        )
         fragment = self.build_fragment(
             template,
-            context,
             initialize_js_func='FreeTextResponseView',
             additional_css=[
                 'public/view.css',
--- a/freetextresponse/tests.py
+++ b/freetextresponse/tests.py
@@ -11,9 +11,10 @@ from opaque_keys.edx.locations import SlashSeparatedCourseKey
 
 from xblock.field_data import DictFieldData
 from xblock.validation import ValidationMessage
+from xblockutils.resources import ResourceLoader
 
 from django.db import IntegrityError
-from django.template.loader import get_template
+from django.template.context import Context
 
 from .freetextresponse import Credit
 from .freetextresponse import FreeTextResponse
@@ -56,7 +57,13 @@ class FreetextResponseXblockTestCase(unittest.TestCase):
         Helper method that creates a Free-text Response XBlock
         """
         course_id = SlashSeparatedCourseKey('foo', 'bar', 'baz')
-        runtime = Mock(course_id=course_id)
+        runtime = Mock(
+            course_id=course_id,
+            service=Mock(
+                # Is there a cleaner mock to the `i18n` service?
+                return_value=Mock(_catalog={}),
+            ),
+        )
         scope_ids = Mock()
         field_data = DictFieldData(kw)
         xblock = FreeTextResponse(runtime, field_data, scope_ids)
@@ -182,10 +189,13 @@ class FreetextResponseXblockTestCase(unittest.TestCase):
         context = {
             'prompt': studio_settings_prompt,
         }
-        template = get_template('freetextresponse_view.html')
+        loader = ResourceLoader('freetextresponse')
+        template = loader.render_django_template(
+            'templates/freetextresponse_view.html',
+            context=Context(context),
+        )
         fragment = self.xblock.build_fragment(
             template,
-            context,
             initialize_js_func='FreeTextResponseView',
             additional_css=[],
             additional_js=[],
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 ddt
 django<1.9,>=1.8
 git+https://github.com/edx/XBlock.git#egg=XBlock
-git+https://github.com/edx/xblock-utils.git@v1.0.0#egg=xblock-utils==v1.0.0
+git+https://github.com/edx/xblock-utils.git@v1.1.1#egg=xblock-utils==v1.1.1
 git+https://github.com/edx/opaque-keys.git@27dc382ea587483b1e3889a3d19cbd90b9023a06#egg=opaque-keys
 -e .
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "xblock-free-text-response",
   "title": "FreeTextResponse XBlock",
   "description": "Enables instructors to create questions with free-text responses.",
-  "version": "0.3.1",
+  "version": "0.4.0",
   "homepage": "https://github.com/Stanford-Online/xblock-free-text-response",
   "author": {
     "name": "Azim Pradhan",
--- a/setup.py
+++ b/setup.py
@@ -27,7 +27,7 @@ class Tox(TestCommand):
 
 setup(
     name="xblock-free-text-response",
-    version="0.3.1",
+    version="0.4.0",
     description="Enables instructors to create questions with free-text responses.",
     license='AGPL-3.0',
     packages=[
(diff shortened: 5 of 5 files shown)


OmarIthawi:
@caesar2164 Could you please take a look?

caesar2164:
Thanks for the PR @OmarIthawi!

I'll be looking at it this sprint (next two weeks).

OmarIthawi:
Thanks @caesar2164! Looking forward for your review :slightly_smiling_face:

aucanpool:
How can I add this commit in my requirements/edx/github.text file?

OmarIthawi:
@aucanpool you can use this branch: https://github.com/OmarIthawi/xblock-free-text-response/tree/omar/hawthorn

aucanpool:
Thank you very much @OmarIthawi. I am using your branch.

freetextresponse/tests.py
```diff
@@ -11,9 +11,10 @@
 
 from xblock.field_data import DictFieldData
 from xblock.validation import ValidationMessage
+from xblockutils.resources import ResourceLoader
 
 from django.db import IntegrityError
-from django.template.loader import get_template
+from django.template.context import Context
```
caesar2164, line 17: Why the switch from `get_template()` to `Context()`?

freetextresponse/freetextresponse.py
```diff
@@ -297,10 +291,15 @@ def student_view(self, context={}):
...
             }
         )
-        template = get_template('freetextresponse_view.html')
+
+        template = self.loader.render_django_template(
+            'public/templates/freetextresponse_view.html',
+            context=Context(context),
+            i18n_service=self.runtime.service(self, "i18n"),
+        )
+
```
caesar2164: let's get rid of extra blank lines

freetextresponse/freetextresponse.py
```diff
@@ -26,16 +25,14 @@
 
 
 @XBlock.needs("i18n")
-class FreeTextResponse(
-    EnforceDueDates,
-    MissingDataFetcherMixin,
-    StudioEditableXBlockMixin,
-    XBlock,
-):
+class FreeTextResponse(EnforceDueDates, MissingDataFetcherMixin, StudioEditableXBlockMixin, XBlock):
```
caesar2164: please keep the arguments in the multiline format (especially with the trailing comma at the end), it makes diffs cleaner and reduces merge conflicts...

caesar2164 (changes requested):
Thanks @OmarIthawi! 

This is looking really good. I just have a few comments/questions.

freetextresponse/freetextresponse.py
stvstnfrd: Specifically here, it doesn't look like anything changed except reformatting the lines?
Especially in that case, we should not make this change as it's not related to the issue at hand.

freetextresponse/freetextresponse.py
```diff
@@ -297,10 +291,15 @@ def student_view(self, context={}):
                 'other_responses': [],
             }
         )
-        template = get_template('freetextresponse_view.html')
+
```
stvstnfrd: Yeah, small style nitpick, but lets remove this empty line here.

freetextresponse/tests.py
stvstnfrd: @caesar2164 I don't think it's switch as much as it's just a deletion of one import (`get_template`) and the addition of another (`Context`).

freetextresponse/tests.py
```diff
@@ -56,7 +57,12 @@ def make_an_xblock(cls, **kw):
         Helper method that creates a Free-text Response XBlock
         """
         course_id = SlashSeparatedCourseKey('foo', 'bar', 'baz')
-        runtime = Mock(course_id=course_id)
+        runtime = Mock(
+            course_id=course_id,
+            service=Mock(
+                return_value=Mock(_catalog={}),
```
stvstnfrd, line 64: What is this used by?

freetextresponse/freetextresponse.py
```diff
@@ -297,10 +291,15 @@ def student_view(self, context={}):
...
+        template = self.loader.render_django_template(
+            'public/templates/freetextresponse_view.html',
+            context=Context(context),
+            i18n_service=self.runtime.service(self, "i18n"),
```
stvstnfrd: Little nitpick, but can you change this to use single quotes? `'i18n'`

freetextresponse/freetextresponse.py
```diff
@@ -2,10 +2,8 @@
 This is the core logic for the Free-text Response XBlock
 """
 from enum import Enum
-import pkg_resources
```
stvstnfrd: This was just unused?

requirements.txt
```diff
@@ -1,6 +1,6 @@
 ddt
 django<1.9,>=1.8
 git+https://github.com/edx/XBlock.git#egg=XBlock
-git+https://github.com/edx/xblock-utils.git@v1.0.0#egg=xblock-utils==v1.0.0
+git+https://github.com/edx/xblock-utils.git@v1.1.1#egg=xblock-utils==v1.1.1
```
stvstnfrd, line 4: What is the impact of this re: ginkgo vs hawthorn, etc.?

tox.ini
```diff
@@ -13,7 +13,7 @@ deps =
 deps =
     -rrequirements.txt
     pep8
-commands = {envbindir}/pep8 freetextresponse/
+commands = {envbindir}/pep8 --max-line-length=120 freetextresponse/
```
stvstnfrd: Which lines necessitated this?

stvstnfrd (changes requested):
Thanks for taking this on!

I have a few questions and a few style nitpicks, but otherwise this is looking good.
I think we'll also need to bump the version number; should be in `setup.py`, I think?

Note: I haven't had a chance to spin this up in a devstack yet, so give me a chance to do that still.

Thanks again and let me know if you have questions/concerns with the feedback!

freetextresponse/freetextresponse.py
```diff
@@ -297,10 +291,15 @@ def student_view(self, context={}):
                 'other_responses': [],
             }
         )
-        template = get_template('freetextresponse_view.html')
+
+        template = self.loader.render_django_template(
+            'public/templates/freetextresponse_view.html',
```
stvstnfrd: Wait, I didn't think the templates were in the `public` directory...

I thought `templates` was a sibling of `public`, isn't it?
Related, do we also need to update `setup.py` to ensure the `templates` directory is packaged,
similar to https://github.com/Stanford-Online/xblock-submit-and-compare/pull/25

OmarIthawi:
@caesar2164 @stvstnfrd Thank you for the review! Will get back to the review soon!

freetextresponse/freetextresponse.py
OmarIthawi: @caesar2164 @stvstnfrd I've used `render_django_template` which is more XBlock friendly, especially for `i18n`.

freetextresponse/tests.py
OmarIthawi: @caesar2164 @stvstnfrd I've used `render_django_template` which is more XBlock friendly, especially for `i18n`.

freetextresponse/freetextresponse.py
OmarIthawi: Oops, moving that out. Thanks @stvstnfrd.

freetextresponse/freetextresponse.py
OmarIthawi: Done

freetextresponse/freetextresponse.py
OmarIthawi: Done

freetextresponse/tests.py
OmarIthawi: So the `i18n` service doesn't freak out. I've added an inline comment to clarify.

tox.ini
OmarIthawi: I must have forgot it. Thanks @stvstnfrd!

requirements.txt
OmarIthawi: Thanks @stvstnfrd. I'll test this one, adding it to the pull request TODO list.

freetextresponse/freetextresponse.py
OmarIthawi: Apparently, yeah.

requirements.txt
OmarIthawi: BTW @stvstnfrd, I've already tested `xblock-utils.git==v1.1.1` on Ficus, and it worked fine. However, I'm happy to re-test in Ficus specifically for this XBlock if you think it's required.

As for Ginkgo, I have no Ginkgo sandbox available to test on, so would appreciate your help.

OmarIthawi:
@caesar2164 @stvstnfrd Thank you for review, I've addressed all of the comments.

I still have to test the updated version in both Hawthorn (just another check) and Ficus.

Please please a look and let me know what do you think.

OmarIthawi:
Good news, it works on both Ficus and Hawthorn. Ticking those boxes in the PR review.

@caesar2164 @stvstnfrd Please take another look.

### Ficus
<kbd>image</kbd>

### Hawthorn
<kbd>image</kbd>

freetextresponse/freetextresponse.py
```diff
@@ -298,10 +297,13 @@ def student_view(self, context={}):
                 'other_responses': self.get_other_answers(),
             }
         )
-        template = get_template('freetextresponse_view.html')
+        template = self.loader.render_django_template(
+            'templates/freetextresponse_view.html',
+            context=Context(context),
+            i18n_service=self.runtime.service(self, 'i18n'),
```
caesar2164, line 303: Hmm, I'm getting an error on our Ginkgo based devstack:

screen shot 2018-12-10 at 11 50 35

What version of xblock-utils are you running in your Ficus devstack? Our Ginkgo stacks are running v1.0.5, and that argument seems to have been added in xblock-utils v1.1.0? (https://github.com/edx/xblock-utils/commit/ba54084a572a618b9248236403c1200f32d4f889)

OmarIthawi: @caesar2164 Looks like I already have `xblock-utils==1.1.1` installed already. We have a couple of other PRs needing i18n and therefore using `xblock-utils==1.1.1`.

How do you suggest moving on with this issue? Should we require updating the xblock-utils or doing a nasty patch to avoid breaking the XBlock on pre-Hawthorn releases.

freetextresponse/tests.py
stvstnfrd: Thanks @OmarIthawi

OmarIthawi:
@caesar2164 I think this PR is ready for merging. 

If upgrading `xblock-utils` is not an option for your fork, there's a workaround, or should I say a hack, in which we can avoid it.

caesar2164 (approved):

caesar2164:
@OmarIthawi, sorry for the hiatus, I forgot to mention that Stanford takes a 2 week "closure" in winter.

On my end, this seems good to go. Only @stvstnfrd's review left.

OmarIthawi:
Thanks @caesar2164! No worries at all, closure is always good :smiley:

requirements.txt
stvstnfrd: Thanks for testing @OmarIthawi . I looked at the relevant commits and this does look like it should be backwards compatible.

freetextresponse/freetextresponse.py
stvstnfrd: @caesar2164 Are you still getting this issue or has it been resolved?

caesar2164: I believe so, looks like @OmarIthawi already pushed the version of xblock-utils to 1.1.1

OmarIthawi:
@stvstnfrd could you please take a look and let me know what do you think?

cc: @caesar2164

stvstnfrd (approved):

caesar2164:
Thanks for the PR @OmarIthawi! Merged!

OmarIthawi:
@caesar2164 @stvstnfrd Thank you for creating this XBlock and for your help here!

PR #40, opened 2019-02-27
Remove package-lock

stvstnfrd:
I don't think we need this any more,
since we aren't using `grunt` or any other NPM packages.


stvstnfrd:
I updated the tests, as  they caused the build to fail.

caesar2164 (commented):
@stvstnfrd could you add a comment saying that the plurals were changed to `AssertEqual` for python 3 compatibility?

stvstnfrd:
@caesar2164 I added a comment and link to documentation in the commit message about the py3 deprecation.

caesar2164 (approved):
doh!

PR #41, opened 2019-04-03
Update dependencies for free-text-response

stvstnfrd:


stvstnfrd:
This is prompted by the security warnings (we don't actually use `requirements.txt` here, so it's a false positive).

kluo (approved):

stvstnfrd:
@kluo btw: I fixed tests; they pass now.

PR #42, opened 2019-07-04
i18n tooling, fixes, more languages plus Transifex hosting

OmarIthawi:
We'd like to add more languages such as Canadian French as a supported language. This is a continuation to https://github.com/Stanford-Online/xblock-free-text-response/pull/35 , I should have completed the work then, but didn't.

Before you continue, take a look at [edx/i18n-tools](https://github.com/edx/i18n-tools) and [edx/cookiecutter-django-app](https://github.com/edx/cookiecutter-django-app).

### Summary of the Changes
 - Switched from `django.utils.translation.ugettext` to `i18n_service.ugettext` to capture what's in the `text.po` files.
 - Added a Makefile to help manage the translations. The file is copied and modified from: [edx/cookiecutter-django-app](https://github.com/edx/cookiecutter-django-app)
 - Added `freetextresponse/locale/config.yaml` which works as the configuration file of the `i18n_tools` command 
 - Refactored the linking of `locale` link to `translations` to streamline it with the current (but undocumented) conventions. See https://github.com/open-craft/xblock-poll/pull/43 for an example of adding full i18n support to an XBlock.
 - Added `requirements-dev.txt` to pin development packages. Used for now only for translations.

### TODO

 - [x] Make Travis happy
 - [x] Push to the [Open edX XBlocks project on Transifex](https://www.transifex.com/open-edx/xblocks/content/)
 - [x] Document how to update the strings properly from Transifex
 - [x] Manually test that the the Arabic translation is working
 - [ ] I would like to pin `pylint`  and others in `requirements-dev.txt`. Unversioned style checker is a mess to follow up with.

### It Works!

<kbd>image</kbd>

PR #43, opened 2019-08-07
Modernize XBlock infrastructure

stvstnfrd:


.tx/config
```diff
@@ -0,0 +1,8 @@
+[main]
+host = https://www.transifex.com
+
+[stanford-openedx-xblocks.free-text-response]
```
OmarIthawi: I've already created a Transifex repository for #42, could we please use that?

 - https://www.transifex.com/open-edx/xblocks/xblock-free-text-response/

This way, when I need to only invite translators (and I do that often) to a single project instead of many Transifex projects.

Makefile
```diff
@@ -0,0 +1,88 @@
...
+	docker build -t $(_NAME) .
+.PHONY: docker_static docker_test docker_translations
+define run-in-docker
+$(docker_make_more) $(patsubst docker_%, %, $@)
+endef
+docker_shell:
+	$(docker_run) --entrypoint /bin/bash $(_NAME)
+docker_static: ; make build_docker; $(run-in-docker)  ## Compile static assets in docker container
+docker_translations: ; make build_docker; $(run-in-docker)  ## Update translation files in docker container
+docker_test: ; make build_docker; $(run-in-docker)  ## Run tests in docker container
```
OmarIthawi, line 107: Thanks for making Docker optional. I'm starting to appreciate the simplicity of a plain-old `virtualenv` inside my machine.

README.rst
```diff
@@ -0,0 +1,125 @@
+Free Text Response XBlock
+==================
```
OmarIthawi:
```suggestion
================================
```

To make it fit under the title.

README.rst
```diff
@@ -0,0 +1,125 @@
...
+~~~~~~~~~~~~~~~~~~~~
+
+To install the XBlock on your platform,
+add the following to your `requirements.txt` file:
+
+    xblock-free-text-response
+
+You'll also need to add this to your `INSTALLED_APPS`:
+
+    freetextresponse
```
OmarIthawi, line 28: Not really. Why would we need this anyway?

freetextresponse/__init__.py
```diff
@@ -3,4 +3,3 @@
 Instructors can specify a list of phrases, of which one must be
 present in order for the student to receive credit.
 """
-from .freetextresponse import FreeTextResponse
```
OmarIthawi, line 6: Awesome!

freetextresponse/models.py
```diff
@@ -0,0 +1,242 @@
+"""
+Handle data access logic for the XBlock
+"""
+from __future__ import absolute_import
+
+from enum import Enum
+from django.db import IntegrityError
+from django.utils.translation import ugettext_lazy as _
```
OmarIthawi, line 8: I really hope I'm wrong about this one, I seriously do. [This issue have wasted a lot of my time for support requests, see this thread in Open edX Discuss](https://discuss.openedx.org/t/translate-the-default-xblock-data/213).

I don't think my dreams would come true immediately, so if you get exceptions regarding that issue, please use the `ugettext_noop` function instead.

```suggestion
from django.utils.translation import ugettext_noop as _
```

freetextresponse/views.py
```diff
@@ -0,0 +1,367 @@
...
+        )
+        return result
+
+    def validate_field_data(self, validation, data):
+        """
+        Validates settings entered by the instructor.
+        """
+        if data.weight < 0:
+            msg = FreeTextResponseViewMixin._generate_validation_message(
+                'Weight Attempts cannot be negative'
```
OmarIthawi, line 329:
```suggestion
                ugettext_lazy('Weight Attempts cannot be negative')
```

right?

freetextresponse/views.py
```diff
@@ -0,0 +1,367 @@
...
+        if data.min_word_count > data.max_word_count:
+            msg = FreeTextResponseViewMixin._generate_validation_message(
+                'Minimum Word Count cannot be greater than Max Word Count'
+            )
+            validation.add(msg)
+        if not data.submitted_message:
+            msg = FreeTextResponseViewMixin._generate_validation_message(
+                'Submission Received Message cannot be blank'
+            )
+            validation.add(msg)
```
OmarIthawi, line 351: Or maybe don't use `ugettext_lazy`, I'm not so sure.

Just bringing your attention to the issue. So, which is better in your opinion?

freetextresponse/xblocks.py
```diff
@@ -0,0 +1,23 @@
+"""
+This is the core logic for the XBlock
+"""
+from __future__ import absolute_import
+from xblock.core import XBlock
+
+from .mixins.scenario import XBlockWorkbenchMixin
```
OmarIthawi, line 7:
```suggestion
from freetextresponse.mixins.scenario import XBlockWorkbenchMixin
```

freetextresponse/xblocks.py
```diff
@@ -0,0 +1,23 @@
...
+from .mixins.user import MissingDataFetcherMixin
+from .models import FreeTextResponseModelMixin
+from .views import FreeTextResponseViewMixin
```
OmarIthawi, line 10: Ditto :point_up:

freetextresponse/locale/vi/LC_MESSAGES/django.po
```diff
@@ -1,48 +0,0 @@
-# Vietnamese Translations for Free Text Response Xblock
-# plural translations are copies of the singular version
-
-# Copyright (C) 2017
-# This file is distributed under the same license as the free-text-response package.
-# Hang Pham <hpham3@stanford.edu>, 2017
-# 
-msgid ""
-msgstr ""
-"Project-Id-Version: 0.1.2\n"
```
OmarIthawi, line 10: Why removing `Vietnamese`?

OmarIthawi (approved):
Thanks @stvstnfrd for making these changes. I've added a couple of questions and amendments, but overall looks awesome.

As a future suggesiton, if you're making more XBlocks, I really recommend the [edx/cookiecutter-xblock](https://github.com/edx/cookiecutter-xblock) it's an awesome tool in my opinion for making new XBlocks.

cc: @caesar2164

freetextresponse/views.py
stvstnfrd: It looks like we're performing translation with `ugettext` inside the helper `_generate_validation_message`,
instead of explicitly translating the returned values from each call.

Does that address your concern?

freetextresponse/views.py
stvstnfrd: See note above: I _think_ we're already translating "properly", assuming that's the concern.

FWIW: I don't think this PR contains any substantive logic changes;
I think this is just showing up in the diff because we renamed/split a few python files.

freetextresponse/xblocks.py
stvstnfrd: Is this a problem or a preference?

I generally prefer relative imports from within the same package;
it's less verbose and is one fewer thing to update if the package is ever renamed
(we've done this a couple time historically w/ xblocks).

freetextresponse/xblocks.py
stvstnfrd: See note above ^

freetextresponse/__init__.py
stvstnfrd: Yeah, we've been removing this superfluous imports from all of our xblocks.

README.rst
stvstnfrd: I think we started adding this when we switched to django templates (instead of simple string interpolation);
we've had it listed in our `INSTALLED_APPS` for a few years now.

Does this render without it?
Happy to remove, if so.

Makefile
stvstnfrd: Yeah, that's the thought.. 

It's nice to run in docker (generally), but I still like having the option to run on bare metal :)

.tx/config
stvstnfrd: Is that repo owned by edx? 
Assuming so, and/or assuming we/Stanford will/can have access easily, maybe we can make it work. I would just want to be sure that we/Stanford can create any new resources that we need there and can easily dole out permissions, otherwise _we_ end up needing to access _two_ repos ;)

It'll probably be better for the community that way and get us more translations (we can probably look to do the same for our other xblocks too).

CC: @caesar2164 @kluo Thoughts?

stvstnfrd (commented):
@OmarIthawi Sorry for the wait. Please take a look when you have some time.

I think I ported over most anything from your PR that wasn't already included here.

Note: There are a few discrepancies between our template and edx's;
we'll maybe work to reconcile that some day, but we're already using this approach though most of our XBlocks now.

freetextresponse/locale/vi/LC_MESSAGES/django.po
stvstnfrd: I think the file is just being renamed to `text.po`?

README.rst
OmarIthawi: XBlock uses templates in a self-contained way, so this is not needed.

Unless there's migration, or platform-wide templates (plus other minor uses) `INSTALLED_APPS` is not needed.

Anyway, we can test it to be safe.

OmarIthawi (approved):
@stvstnfrd et. al. Thank you for this work. I quickly checked the changes and they look good to me:

 - :heavy_check_mark: Using the community Transifex resource.
 - :heavy_check_mark: `i18n_service` is being used correctly -- I guess :sweat_smile: 

Generally if it works, then let's go with it. If it doesn't, someone's (you or I) going to fix it.

.tx/config
OmarIthawi: Yes it's owned by edX and I'm also an admin so ping me if you needed anything.

We also would like to democratize it as much as possible so the Translators don't have to login into a dozen different account.

freetextresponse/views.py
OmarIthawi: Nope, as long as it works it's not an issue.

freetextresponse/xblocks.py
OmarIthawi: Not a problem, but I've "seen it somewhere in a linter" type of preferences.

caesar2164 (approved):
I have nothing to add to this, so 👍

